PR #7, opened 2020-11-16
Add multithreaded tensor preparation on C++ side and binpack loader.

Sopel97:
This PR adds multithreaded data preparation on the C++ side from which both bin and binpack benefit, but bin has other bottlenecks. It also introduces a multithreaded binpack reader which is used by the data preparation function.

There is a new option `--num-workers` which specifies the number of worker threads to use for data preparation. The threads are assigned to feature gathering and binpack reading in 2:1 ratio (or as close to that ratio as possible).

Please verify if this works correctly on your end before commiting.


vondele:
warning:
``` 
nnue-pytorch/lib/nnue_training_data_formats.h:7677:46: warning: control reaches end of non-void function [-Wreturn-type]
 7677 |             std::unique_lock lock(m_fileMutex);
```
assert
```
nnue-pytorch/lib/nnue_training_data_formats.h:6773: binpack::CompressedTrainingDataFile::Header binpack::CompressedTrainingDataFile::readChunkHeader(): Assertion `false' failed.
```

vondele:
both with 1 and 4 num-workers.

Sopel97:
The second error is probably from UB guided optimization.
I fixed the first issue.

vondele:
warning and crash fixed!

GPU load and speed looks the same here (num-workers 4)

vondele:
The first epoch of the training run looks similar, so I think it is working correctly.

vondele:
The benchmark runs like this for me:
```
1 11.6234s
2 11.6578s
4 7.47875s
8 10.2587s
```

vondele:
If I read the benchmark correctly, it is doing 10000 * 8192 positions in ~10s that's 8M/s.... in that case, can't be a bottleneck in the trainer.

vondele:
hmm. I have a training run that seems stuck after 4 epochs. Python process alive, but no GPU load. I'll retry a run, just wondering if this could be related to this patch, since it is the first time I have it, and maybe this could deadlock?

vondele:
happened a second time (now epoch 5). I'll do an additional experiment since I also used 
`--progress_bar_refresh_rate 0` (but I'm watching the logs/ directory).

Sopel97:
There is one potential race that I see. I'll try to redo this part differently.

vondele:
I had an additional hang, and it happened when the validation loop started.  So basically, I think we're reading training and validation data at the same time (not sure).

vondele:
an other hang, also with training and validation at the same time.

Sopel97:
I think the possible race condition was due to having `hasNext` type iterable, which when combined with worker threads is hard to get correct. I've changed the design around that so that it always terminates, not only when `hasNext` would be true. Also I identified one lock contention and improved the situation there, and some other minor changes. I've also fixed a race that resulted in the data loader exiting when next() was called before threads actually started.

I've done 15 epochs with parallel validation, each epoch 1M positions, 4 workers. I experienced no issues.

vondele:
I'll be testing a little more.

vondele:
I think it is fixed, currently running fine at epoch 9, better than any of the 5 runs before.

PR #9, opened 2020-11-18
Add some additional arguments and output

vondele:
show filename of training and validation data

allow to set the number of torch threads
allow to set the batch size


vondele:
some additional output I was missing looking at the logs of my runs, and two flags that are useful to tune performance.

glinscott:
LGTM, thanks!

PR #11, opened 2020-11-19
Additional command line arguments

vondele:
--default_root_dir to specify a non-default log dir
--seed to specify a user controlled seed


glinscott:
LGTM!

PR #12, opened 2020-11-21
Add multigpu support.

Sopel97:
This is a WIP PR to add multigpu (or rather, multidevice) support. Currently it's in a crude state and is not fully configurable through cli options. `--gpus` must NOT be passed, instead one has to modify the `devices = ['cuda:0']` in `train.py`. The code should work on multiple devices as it is. The main purpose of this draft is to ensure that this WORKS at all and that the results are correct. The performance when using one device should be unchanged. The performance when using multiple devices may not be good, this will be improved once we know this is working correctly.

To make this work with pytorch lightning I had to slightly abuse the `*_step_end` functions and override the `backward()` function with an empty one. Hopefully that's fine.

The code is based on my initial toy example here https://github.com/Sopel97/nnue-pytorch/blob/multigpu_research/test.py

@vondele I'd like to get your feedback as you're the only person I know who has multiple gpus to try this code. I've verified that this works in a cpu + gpu configuration, but I'd like to confirm this also works on multigpu configuration and possibly get some feedback regarding performance.

I would also like to get some feedback on some other stuff regarding this.
1. `--gpus` becomes unusable and MUST not be specified, otherwise pytorch lightning breaks things. Do we always do `args.gpus = None` and issue a warning if it's set?
2. How do we specify the devices used? This is not only limited to gpus with this code. I was thinking of `--devices=[cpu,cuda:0,cuda:1,whatever]`. Possibly take over `--gpus` parameter, but I'm not keen on parsing all possible syntaxes this parameter accepts in pytorch.
3. What do we do with the python data loader? I don't really know from the top of my head how to make it produce multiple batches, given that it uses pytorch DataLoader, and I don't feel like spending time on it.

### DIFF
--- a/model.py
+++ b/model.py
@@ -20,20 +20,24 @@ class NNUE(pl.LightningModule):
 
   It is not ideal for training a Pytorch quantized model directly.
   """
-  def __init__(self, feature_set=halfkp, lambda_=1.0):
+  def __init__(self, feature_set=halfkp, lambda_=1.0, devices=['cpu']):
     super(NNUE, self).__init__()
     num_inputs = feature_set.INPUTS
-    self.input = nn.Linear(num_inputs, L1)
+
+    self.devices = devices
+    self.main_device = devices[0]
+
+    self.input = nn.Linear(num_inputs, L1).to(device=self.main_device)
 
     # Zero out the weights/biases for the factorized features
     # Weights stored as [256][41024]
     weights = self.input.weight.narrow(1, 0, feature_set.INPUTS - feature_set.FACTOR_INPUTS)
-    weights = torch.cat((weights, torch.zeros(L1, feature_set.FACTOR_INPUTS)), dim=1)
+    weights = torch.cat((weights, torch.zeros(L1, feature_set.FACTOR_INPUTS, device=self.main_device)), dim=1)
     self.input.weight = nn.Parameter(weights)
 
-    self.l1 = nn.Linear(2 * L1, L2)
-    self.l2 = nn.Linear(L2, L3)
-    self.output = nn.Linear(L3, 1)
+    self.l1 = nn.Linear(2 * L1, L2).to(device=self.main_device)
+    self.l2 = nn.Linear(L2, L3).to(device=self.main_device)
+    self.output = nn.Linear(L3, 1).to(device=self.main_device)
     self.lambda_ = lambda_
 
   def forward(self, us, them, w_in, b_in):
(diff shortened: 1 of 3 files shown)


vondele:
crashes for me:
```
$ python train.py ../nnue-pytorch/data04_2B.binpack ../nnue-pytorch/d10_128000_6293.binpack 
Ranger optimizer loaded. 
Gradient Centralization usage = True
GC applied to both conv and fc layers
Training with ../nnue-pytorch/data04_2B.binpack validating with ../nnue-pytorch/d10_128000_6293.binpack
Seed 42
Using batch size 8192.0
Using c++ data loader
  File "train.py", line 81, in <module>
    main()
  File "train.py", line 71, in main
    train, val = data_loader_cc(args.train, args.val, args.num_workers, batch_size, devices=devices)
  File "train.py", line 19, in data_loader_cc
    train = DataLoader(nnue_dataset.FixedNumBatchesDataset(train_infinite, (epoch_size + batch_size - 1) // batch_size), batch_size=None, batch_sampler=None)
  File "/users/vjoost/test/nnue-pytorch-exp2/nnue_dataset.py", line 140, in __init__
    self.iter = iter(self.dataset)
  File "/users/vjoost/test/nnue-pytorch-exp2/nnue_dataset.py", line 134, in __iter__
    devices=self.devices)
  File "/users/vjoost/test/nnue-pytorch-exp2/nnue_dataset.py", line 115, in __init__
    devices)
  File "/users/vjoost/test/nnue-pytorch-exp2/nnue_dataset.py", line 68, in __init__
    self.stream = self.create_stream(self.feature_set, self.num_workers, self.filename, batch_size, cyclic)
ctypes.ArgumentError: argument 4: <class 'TypeError'>: Don't know how to convert parameter 4
Exception ignored in: <function TrainingDataProvider.__del__ at 0x7ffef4f6ae60>
  File "/users/vjoost/test/nnue-pytorch-exp2/nnue_dataset.py", line 91, in __del__
    self.destroy_stream(self.stream)
AttributeError: 'SparseBatchProvider' object has no attribute 'stream'

```

Sopel97:
So the issue right now seems to be that (at least) backward synchronizes with the host device in each loop over loss.
our test.py that worked well
```
backward  0.004 seconds
sync      0.172 seconds
```
nnue trainer
```
backward  0.056 seconds
sync      0.010 seconds
```

Sopel97:
https://discuss.pytorch.org/t/long-time-spent-in-backward-when-inputs-are-sparse-tensors/103695

Sopel97:
As we received no answers this PR looks not pursuable. I'll close it.

PR #15, opened 2020-11-23
Refactor filtering.

Sopel97:
Refactor position skipping to be easier to configure, more centralize…d, togglable by cli argument, and work for .bin. Add --smart-fen-skipping option (flag) that mimicks nodchip trainer option with the same name, default to false.

For parallel binpack loader the filtering happens in the worker threads during reading the binpack file. For non-parallel binpack reader and for bin readers the filtering happens at stream level. The predicate for skipping is configurable before stream construction. Filtering is dynamically togglable from python side.

Note that the python data loader still does not handle this option.


glinscott:
Thanks!

PR #21, opened 2020-12-04
Cleanup model import/export with factorizer

glinscott:
@vondele @Sopel97 

train.py
- Adds `--enable-factorizer` option.  Default is false, which will train without factorizer.  Factorizer on or off dynamically modifies the model shape, so you can resume non-factorized nets in factorizer mode.
- Adds `--reuse-from-model` option.  Pass in a serialized model, ending in `.pt`, to resume from those weights.

serialize.py: 
- Fixes nnue -> pt, as they need to be imported without factorizer shapes.
- Supports both factorized and non-factorized pt -> nnue conversions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,11 @@ Import an existing SF NNUE network to the pytorch network format.
 python serialize.py nn.nnue converted.pt
 ```
 
+Import the network, with the factorizer enabled (for resuming training):
+```
+python serialize.py nn.nnue converted.pt --enable-factorizer
+```
+
 # Logging
 
 ```
--- a/README.md
+++ b/README.md
@@ -35,6 +35,11 @@ python train.py --resume_from_checkpoint <path> ...
 python train.py --gpus 1 ...
 ```
 
+## Enable factorizer
+```
+python train.py --enable-factorizer ...
+```
+
 # Export a network
 
 Using either a checkpoint (`.ckpt`) or serialized model (`.pt`),
@@ -49,11 +54,6 @@ python serialize.py last.ckpt nn.nnue
 Import an existing SF NNUE network to the pytorch network format.
 ```
 python serialize.py nn.nnue converted.pt
-```
-
-Import the network, with the factorizer enabled (for resuming training):
-```
-python serialize.py nn.nnue converted.pt --enable-factorizer
 ```
 
 # Logging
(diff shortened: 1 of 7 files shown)


Sopel97:
would make sense to have the option be `--enable-factorizer` for both?

halfkp.py
```diff
@@ -27,7 +29,7 @@ def halfkp_idx(is_white_pov: bool, king_sq: int, sq: int, p: chess.Piece):
 def get_halfkp_indices(board: chess.Board):
   # TODO - this doesn't support the factors yet.
   def piece_indices(turn):
-    indices = torch.zeros(INPUTS)
+    indices = torch.zeros(Features.inputs)
     for sq, p in board.piece_map().items():
       if p.piece_type == chess.KING:
         continue
```
Sopel97: perhaps these should now be a part of `class Features:`

Sopel97: also, the lines https://github.com/glinscott/nnue-pytorch/blob/7793bf6620fcd8003692d755c0e735f8ff5566de/serialize.py#L12-L32 could become a part of the factorizer

glinscott: Good call, done.

glinscott:
Switched both options to `--enable-factorizer`.

glinscott:
Actually, ended up removing `--enable-factorizer` from serialize.py.  It's cleaner to only have it in train.py, which now adjusts the model based on the options passed into training.

README.md
```diff
@@ -51,6 +51,11 @@ Import an existing SF NNUE network to the pytorch network format.
 python serialize.py nn.nnue converted.pt
 ```
 
+Import the network, with the factorizer enabled (for resuming training):
+```
+python serialize.py nn.nnue converted.pt --enable-factorizer
```
vondele: needs update?

glinscott: Done, thanks.

glinscott:
Ok, I think this is ready to merge now if it looks good to both of you.

vondele:
I've just been testing and either I'm doing something wrong, or there is an issue:
```
python serialize.py nn-c3ca321c51c9.nnue nn-c3ca321c51c9.pt
python train.py  --resume-from-model nn-c3ca321c51c9.pt --gpus 1 ../gensfen/data05/d5_128000000_1468_staticEval.binpack ../gensfen/data05/d5_128000000_1468_staticEval.binpack
```
gives a large loss: loss=0.060 whereas it should be 0.000

glinscott:
Oops, good catch.  Pushed a fix.

vondele:
excellent restarts from .pt now work with and without factorizer, giving the expected zero loss.

PR #23, opened 2020-12-04
Pass feature counts and remove the magic `20` constant for halfk factorizer.

Sopel97:
This PR adds value arrays in the sparse batch that stand for the feature count. Now all features for a single position should have unique indices, but may have different counts (weights).

This is untested, please verify that there are no obvious mistakes.

### DIFF
--- a/training_data_loader.cpp
+++ b/training_data_loader.cpp
@@ -56,7 +56,7 @@ struct HalfKP {
         return 1 + static_cast<int>(orient(color, sq)) + p_idx * NUM_SQ + static_cast<int>(ksq) * NUM_PLANES;
     }
 
-    static int fill_features_sparse(int i, const TrainingDataEntry& e, int* features, int& counter, Color color)
+    static int fill_features_sparse(int i, const TrainingDataEntry& e, int* features, float* values, int& counter, Color color)
     {
         auto& pos = e.pos;
         auto pieces = pos.piecesBB() & ~(pos.piecesBB(Piece(PieceType::King, Color::White)) | pos.piecesBB(Piece(PieceType::King, Color::Black)));
@@ -65,9 +65,10 @@ struct HalfKP {
         {
             auto p = pos.pieceAt(sq);
             int idx = counter * 2;
-            counter += 1;
             features[idx] = i;
             features[idx + 1] = feature_index(color, orient(color, ksq), sq, p);
+            values[counter] = 1.0f;
+            counter += 1;
         }
         return INPUTS;
     }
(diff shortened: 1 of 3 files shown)


vondele:
at least the --resume-from-model with a static binpack gives the expected 0.000 loss, also with factorizer.

glinscott:
We will need to do some ELO testing of the exported nets though, as we need to test that the weights are set consistently with the export.  I will be able to test this out tomorrow.  @vondele if you beat me to it, I was just going to run a factorizer enabled net through serialize.py, and compare elo against the original.

training_data_loader.cpp
```diff
@@ -83,17 +84,20 @@ struct HalfKPFactorized {
...
         auto& pos = e.pos;
         {
+            auto num_added_features = counter - counter_before;
             // king square factor
             auto ksq = pos.kingSquare(color);
             int idx = counter * 2;
-            counter += 1;
             features[idx] = i;
             features[idx + 1] = offset + static_cast<int>(orient(color, ksq));
+            values[counter] = static_cast<float>(num_added_features);
```
glinscott, line 99: This works because we don't include the kings as pieces, but perhaps calling the variable `num_non_king_pieces` or something might be a bit clearer?

glinscott:
Code changes LGTM otherwise.

vondele:
@glinscott I'll do a test of this branch now.
1) serialize a .nnue
2) restart from that .pt
3) optimize a couple of epochs
4) Elo test the resulting net.

training_data_loader.cpp
Sopel97: Even if the (king king) features are included (as in halfka) each such (king king) feature can be factorized the same as all others, i.e. halfk * p. So this value is not `num_non_king_pieces`

vondele:
so done a comparison between patch and master. Generally, I would say pretty similar. Exported nets with similar Elo performance -30Elo and -25Elo afte about 8 epochs after a restart from master. Loss very slightly lower with master, compared to patch, master also 5Elo better, but that's roughly the noise. Also we don't know if either is just better converging away from the initial master net. 

For the sake of consistency, I think it makes sense to merge it. But that just my 2cents.

vondele:
image

training loss comparison.

Sopel97:
Should we try training a net without halfk factors at all before merging this?

vondele:
you mean same procedure but without '--enable-factorizer ' ? Only if it makes sense, this is not a quick check.

Sopel97:
I mean only include the P factors. So now it's 
```
  factors = {
    'kings': 64,
    'pieces': 640,
  }
```
What I would like to try is
```
  factors = {
    'pieces': 640,
  }
```
this obviously requires some changes in a few places.

vondele:
OK, so different experiment in my mind. Needs to be done with more care. I think the good property of this patch is no magic constants. After that we can try to find the best factorizer.

training_data_loader.cpp
glinscott: I see, yes, that's a good point.

PR #24, opened 2020-12-05
avoid python 3.8 math.prod

vondele:


vondele:
with this change I can more easily use the code on a system with python 3.7

Unless we want to require a certain minimum python level

Sopel97:
I think we should strive for the lowest python version possible. This change is minimal, everything works.

PR #25, opened 2020-12-05
Test for quantization error by comparing evals.

Sopel97:
This adds an option to test quantization error during serialization/deserialization. Currently it doesn't properly support factorized nets so only works well on deserialization - to make it work with factorized nets we would have to first unfactorize them.

Example usage:
```
python serialize.py nn.nnue a.pt --test-eval-error --test-data d8_100000.binpack --test-num-samples 2048
```

Example output:
```
Avg abs eval:      18.98697853088379
Avg abs eval diff: 1.4802887439727783
Relative error:    0.07798033952713013
```

I'm not sure if the quantized model is 100% correct. I'll leave this as a draft so that this functionality is accessible for testing.

### DIFF
--- a/model.py
+++ b/model.py
@@ -11,6 +11,35 @@ L1 = 256
 L2 = 32
 L3 = 32
 
+class QuantizedNNUE:
+  def __init__(self, model, feature_set=halfkp.Features()):
+    self.feature_set = feature_set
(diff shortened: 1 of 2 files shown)


Sopel97:
Something seems to be not correct. nnue -> pt should yield 0 relative difference, but its about 0.5% for me.

PR #27, opened 2020-12-06
Add --random-fen-skipping N

vondele:
--random-fen-skipping 5

Will skip on average 5 fens before accepting one.
With very large binpacks allows for more diversity in a batch/epoch, not using all moves of a game in that batch.
Randomness ensures that on repeated runs (or cycling) on the same binpack, all fens will eventually be used.

Little slowdown, especially with using --threads N --num-workers N,
even for large skipping values (e.g. skipping 10 -> 10% slowdown)

### DIFF
--- a/training_data_loader.cpp
+++ b/training_data_loader.cpp
@@ -7,6 +7,7 @@
 #include <mutex>
 #include <thread>
 #include <deque>
+#include <random>
 
 #include "lib/nnue_training_data_formats.h"
 #include "lib/nnue_training_data_stream.h"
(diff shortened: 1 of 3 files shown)


training_data_loader.cpp
```diff
@@ -396,13 +397,16 @@ struct FeaturedBatchStream : Stream<StorageT>
...
         std::function<bool(const TrainingDataEntry&)> skipPredicate = nullptr;
-        if (filtered)
+        if (filtered || random_fen_skipping)
         {
-            skipPredicate = [](const TrainingDataEntry& e){
-                return e.isCapturingMove() || e.isInCheck();
+            skipPredicate = [random_fen_skipping, filtered](const TrainingDataEntry& e){
+                std::random_device rd;
+                static thread_local std::mt19937 gen(rd()) ;
+                std::uniform_int_distribution<> distrib(0, random_fen_skipping);
```
Sopel97: could use bernoulli distribution and pass a float?

vondele: OK, done.

Sopel97:
If the performance is not degraded when combined with the shuffle patch (because shuffling has higher priority in my mind) I can merge this.

vondele:
not really, of course, when skipping many fens it becomes slower.

I'm locally testing both patches combined, and loss curves look better to the eye (i.e. noisy around the smoothed average smaller, maybe somewhat faster convergence, slightly difficult to see as tensorboard 'sparsifies' the curves, it seems):
image

PR #28, opened 2020-12-06
shuffle fen reading buffer before use

vondele:
shuffles the fens before use, increase the buffer used for reading

speed essentially unchanged (18.9 -> 18.8 ?) and slight increase in resident memory (3.5 -> 4.2G)

### DIFF
--- a/lib/nnue_training_data_formats.h
+++ b/lib/nnue_training_data_formats.h
@@ -44,6 +44,7 @@ THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #include <optional>
 #include <thread>
 #include <mutex>
+#include <random>
 
 #if (defined(_MSC_VER) || defined(__INTEL_COMPILER)) && !defined(__clang__)
 #include <intrin.h>
@@ -7577,6 +7578,10 @@ namespace binpack
 
                     if (!m_localBuffer.empty())
                     {
+                        // now shuffle the local buffer
+			static thread_local std::mt19937 g(std::random_device{}());
+			std::shuffle(m_localBuffer.begin(), m_localBuffer.end(), g);
+
                         std::unique_lock lock(m_waitingBufferMutex);
                         m_waitingBufferEmpty.wait(lock, [this]() { return m_waitingBuffer.empty() || m_stopFlag.load(); });
                         m_waitingBuffer.swap(m_localBuffer);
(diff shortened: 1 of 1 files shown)


lib/nnue_training_data_formats.h
```diff
@@ -7577,6 +7578,11 @@ namespace binpack
 
                     if (!m_localBuffer.empty())
                     {
+                        // now shuffle the local buffer
+			std::random_device rd;
+			static thread_local std::mt19937 g(rd());
```
Sopel97: do static thread_local std::mt19937 g(std::random_device()); so that the random device it's created once

vondele: doesn't seem allowed ?:
`/home/vondele/chess/vondele/nnue-pytorch/lib/nnue_training_data_formats.h: In lambda function:
/home/vondele/chess/vondele/nnue-pytorch/lib/nnue_training_data_formats.h:7582:11: error: storage class ‘thread_local’ invalid for function ‘g’
 7582 |    static thread_local std::mt19937 g(std::random_device());
      |           ^~~~~~~~~~~~
/home/vondele/chess/vondele/nnue-pytorch/lib/nnue_training_data_formats.h:7582:4: error: cannot declare static function inside another function
 7582 |    static thread_local std::mt19937 g(std::random_device());
      |    ^~~~~~
`

lib/nnue_training_data_formats.h
```diff
@@ -7682,7 +7688,7 @@ namespace binpack
         CompressedTrainingDataFile m_inputFile;
         std::atomic_int m_numRunningWorkers;
 
-        static constexpr int threadBufferSize = 256 * 256 * 4;
+        static constexpr int threadBufferSize = 256 * 256 * 16;
```
Sopel97, line 7690: IIRC the size of a single entry is ~200B, that means each reading thread will take about 200MB with this. I guess this is fine

lib/nnue_training_data_formats.h
Sopel97: do static thread_local std::mt19937 g(std::random_device()); so that the random device it's created once

Sopel97: right, `static thread_local std::mt19937 g(std::random_device{}());`

vondele: done

lib/nnue_training_data_formats.h
vondele: yes, I think that's fine.

lib/nnue_training_data_formats.h
```diff
@@ -7577,6 +7578,10 @@ namespace binpack
 
                     if (!m_localBuffer.empty())
                     {
+                        // now shuffle the local buffer
+			static thread_local std::mt19937 g(std::random_device{}());
```
Sopel97, line 7582: Usually mt199937_64 is faster and we don't really target 32 bit systems. If the currently version is as fast for you as master I can merge as is.

vondele: yes, no noticeable slowdown.

Sopel97:
I'll just note that this solution only works for .binpack so if we ever want to support .bin seriously we need to do shuffling there too.

PR #29, opened 2020-12-06
Only instantiate a single rng per thread. More laziness in skipPredicate.

Sopel97:
This should be non-function apart from the rng change. This ensures only one rng per thread is created. Also the skipPredicate doesn't instantiate variables related to random skipping if it's not used.


glinscott (approved):

glinscott:
LGTM

PR #30, opened 2020-12-07
A better abstraction for features on the python side.

Sopel97:
This PR introduces a more robust and extensible abstraction for features (for now only on the python side).

A feature transformer is defined over a `feature_set`. A `feature_set` is a list of `feature_block`s. Each `feature_block` contains a name, hash, num_real_features, num_virtual_features, num_features, and knows some stuff about how to manipulate its indices. The number of real features and virtual features is used to distinguish factorizers, because they have to be handled differently on serialization to .nnue. A `feature_block` consists of `factors` which are an `OrderedDict` of `(str, int)` pairs where the first is the name of the factor and the second is its size. The first factor is assumed to be real, all the following ones are assumed to be virtual. `feature_set` concatenates these blocks and exposes similar operations. A `feature_block` knows how to coalesce its feature to real features. A `feature_block` is identified by it's name, for example `HalfKP` is a block, as is `HalfKP^` which denotes a factorized `HalfKP` (this is just a convention, but is strict and is used in some places for conversion ability discovery). A `feature_set` name is `'+'.join(...)` of names of its blocks.

From now on the feature set used for learning and serialization/deserialization has to be specified explicitly as a program argument. The new argument `--features=...` takes a name of the feature set. For example `--features="HalfKP^"` or `--features="HalfKP"` or some imaginary `--features="HalfKP^+FancyFeatures+MoreFancyFeatures^"`. This argument is present in both train.py and serialize.py.

The current semantics are as follows:
1. When training a new net from scratch - `--features` specifies the feature set to use for learning. The feature transformer weights are initialized normally for the real features and zero initialized for the virtual features.
2. When resuming training from a .pt model - `--features` specifies the feature set to use for learning. If the feature set specified doesn't match the feature set from the .pt model a conversion is attempted. Right now only a conversion of feature set with a single block from non-factorized to factorized is supported. The factorized block must have the non-factorized features as the first factor. The virtual feature weights are initialized to zero.
3. When converting .ckpt to .nnue - `--features` specifies the features as stored in the .ckpt file. `load_from_checkpoint` doesn't correctly call the constructor with correct feature_set. I have not explored this and it may be improved, I don't know exactly how this works. The weights are coalesced when writing the .nnue file.
4. When converting .pt to .nnue - `--features` is ignored, the `feature_set` from the saved model is used, the weights are coalesced when writing the .nnue file.
5. When converting .nnue to .pt - `--features` specifies the features in the .nnue file. The resulting .pt model has the same feature_set. Note that when resuming training this model can be converted to a compatible feature_set, see point 2.

Known defects:
- the description string for the net is not dependent on the feature set used. This is a minor issue because it's ignored when loading the net.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -1,0 +1,15 @@
+# Features
+
+This doc is pretty much a copy-paste from [here](https://github.com/glinscott/nnue-pytorch/pull/30)
+
+A feature transformer is defined over a `feature_set`. A `feature_set` is a list of `feature_block`s. Each `feature_block` contains a name, hash, num_real_features, num_virtual_features, num_features, and knows some stuff about how to manipulate its indices. The number of real features and virtual features is used to distinguish factorizers, because they have to be handled differently on serialization to .nnue. A `feature_block` consists of `factors` which are an `OrderedDict` of `(str, int)` pairs where the first is the name of the factor and the second is its size. The first factor is assumed to be real, all the following ones are assumed to be virtual. `feature_set` concatenates these blocks and exposes similar operations. A `feature_block` knows how to coalesce its feature to real features. A `feature_block` is identified by it's name, for example `HalfKP` is a block, as is `HalfKP^` which denotes a factorized `HalfKP` (this is just a convention, but is strict and is used in some places for conversion ability discovery). A `feature_set` name is `'+'.join(...)` of names of its blocks.
+
+From now on the feature set used for learning and serialization/deserialization has to be specified explicitly as a program argument. The new argument `--features=...` takes a name of the feature set. For example `--features="HalfKP^"` or `--features="HalfKP"` or some imaginary `--features="HalfKP^+FancyFeatures+MoreFancyFeatures^"`. This argument is present in both train.py and serialize.py.
+
+The current semantics are as follows:
+
+1. When training a new net from scratch - `--features` specifies the feature set to use for learning. The feature transformer weights are initialized normally for the real features and zero initialized for the virtual features.
+2. When resuming training from a .pt model - `--features` specifies the feature set to use for learning. If the feature set specified doesn't match the feature set from the .pt model a conversion is attempted. Right now only a conversion of feature set with a single block from non-factorized to factorized is supported. The factorized block must have the non-factorized features as the first factor. The virtual feature weights are initialized to zero.
+3. When converting .ckpt to .nnue - `--features` specifies the features as stored in the .ckpt file. The user must pass the correct feature set through `--features` because it can't be inferred from the .ckpt. If the features from `--features` and the saved model don't match it'll likely stack trace on some dimension mismatch.
+4. When converting .pt to .nnue - `--features` is ignored, the `feature_set` from the saved model is used, the weights are coalesced when writing the .nnue file.
+5. When converting .nnue to .pt - `--features` specifies the features in the .nnue file. The resulting .pt model has the same feature_set. Note that when resuming training this model can be converted to a compatible feature_set, see point 2.
--- a/feature_block.py
+++ b/feature_block.py
@@ -4,6 +4,49 @@ def _get_main_factor_name(full_name):
     return full_name.replace('^', '')
 
 class FeatureBlock:
+    '''
+    This is the base class for all the network input features.
+    All features must inherit from this class.
+    It abstracts a named set of features in a way that
+    allows seamless introduction of factorizers.
+
+    For example a set of HalfKP features is a subclass of this class, and
+    so is the HalfKP^ feature set (where "^" denotes that it's factorized).
+
+    There are 3 fundamental things about a feature block which it needs for construction:
+        - name - whatever, just please use ascii. Also we'd like "^" to be reserved to
+                 denote that the block is a factorized version of some other block.
+        - hash - a 32 bit unsigned integer, as defined in the original nodchip trainer
+        - factors - the ordered list of named "festure subblocks". If there's more than one
+                    it's assumed that it's a factorized feature block.
+
+    More about factors, because it's the fundamental building block.
+    A block can have just one factor, like HalfKP, but sometimes it's possible to
+    factorize some features further. Ideally we don't want to have multiple
+    features talking about the same thing when the net is actually used for play,
+    because it's wasteful, but it's helpful during training because it makes it
+    easier to generalize over similar positions. This is for example utilized by HalfKP^,
+    which defines 3 factors: HalfKP, HalfK, and P.
+    Factors are passed to the constructor as an OrderedDict from string to the number of dimensions.
+    The first factor is the "real" factor (or "main" factor), one that is supposed to be used for play.
+    The following factors (if any) are the "virtual" factors, and are only used for training.
+    Based on these factors and their dimensions FeatureBlock defines 3 values:
+        - num_real_features - the number of unfactorized features that the resulting net will use in play
+        - num_virtual_features - the number of additional features used for learning
+                                 that will be coalesced when converting to .nnue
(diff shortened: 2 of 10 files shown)


train.py
```diff
@@ -43,23 +44,23 @@ def main():
   parser.add_argument("--random-fen-skipping", default=0, type=int, dest='random_fen_skipping', help="skip fens randomly on average random_fen_skipping before using one.")
   parser.add_argument("--enable-factorizer", dest='enable_factorizer', action='store_true', help="Enables using the factorizer for training.")
```
vondele: is this option still needed?

features.py
```diff
@@ -0,0 +1,32 @@
...
+def get_feature_set_from_name(name):
+    feature_block_names = name.split('+')
+    blocks = get_feature_blocks_from_names(feature_block_names)
+    return FeatureSet(blocks)
+
+def get_available_feature_blocks_names():
+    return list(iter(_feature_blocks_by_name))
+
+def add_argparse_args(parser):
+    parser.add_argument("--features", dest='features', help="The feature set to use. Can be a union of feature blocks (for example P+HalfKP). \"^\" denotes a factorized block. Currently available feature blocks are: " + ', '.join(get_available_feature_blocks_names()))
```
vondele: maybe have a default "HalfKP^" for now?

vondele:
I assume that's the basis for adding new features to the trainer, which of course would be useful.

I can't really comment on the implementation... just some comments on usability..

I'm not sure I can follow your point 3. Are we talking about --resume_from_checkpoint ? I'm using that option a lot, so would be convenient if that would remain in working state (using the same --features before and after resume)

train.py
Sopel97: right, it should be removed.

Sopel97:
The point 3. is about conversion from nnue to ckpt. The --resume_from_checkpoint is covered in point 2. Resuming from checkpoint can work either on the architecture of the saved model or with an added factorizer. So you can convert .nnue to .pt and train it with a factorizer by specifying `--features="HalfKP^"`

vondele:
maybe I'm confused by the naming a checkpoint is a `.ckpt` for me, a model is a `.pt ` I have been using the following paths

with the serializer:
`.ckpt` -> `.nnue`
`.nnue` -> `.pt`
you talk about 'conversion from nnue to ckpt' which I don't think is possible.

with the trainer : 
`--resume-from-model foo.pt`
`--resume_from_checkpoint bar.ckpt`

now I realize that you are talking about the python function `load_from_checkpoint` which is not related to the command line options.

Sopel97:
Yea sorry, I meant ckpt -> nnue but got all mixed up. Good you figured it out either way.

feature_block.py
```diff
@@ -0,0 +1,32 @@
+from collections import OrderedDict
+
+def _get_main_factor_name(full_name):
+    return full_name.replace('^', '')
+
+class FeatureBlock:
```
glinscott, line 6: Comment about what class is used for?

features.py
```diff
@@ -0,0 +1,42 @@
+from feature_block import *
+from feature_set import *
+
+import argparse
+
+import halfkp
+
+_feature_modules = [halfkp]
```
glinscott, line 14: Is this where you would add new architectures?  If so, a description here would be great.

halfkp.py
```diff
@@ -12,58 +15,35 @@ def halfkp_idx(is_white_pov: bool, king_sq: int, sq: int, p: chess.Piece):
...
-      result.append(w)
-    return torch.cat(result, dim=1)
+  def get_feature_factors(self, idx):
+    if idx >= self.num_real_features:
+      raise Exception('Feature must be real')
 
+    k_idx = idx // NUM_PLANES
+    p_idx = idx % NUM_PLANES - 1
+
+    return [idx, self.get_factor_base_feature('HalfK') + k_idx, self.get_factor_base_feature('P') + p_idx]
```
glinscott, line 46: Nice!

model.py
```diff
@@ -20,31 +19,43 @@ class NNUE(pl.LightningModule):
...
-      self.input = nn.Linear(num_inputs, L1)
-
-    if self.factorizer is not None:
-      # Zero out the weights/biases for the factorized features
-      # Weights stored as [256][41024]
-      weights = self.input.weight.narrow(1, 0, self.feature_set.inputs)
-      weights = torch.cat((weights, torch.zeros(L1, self.factorizer.inputs)), dim=1)
+    self._zero_virtual_feature_weights()
+
+  def _zero_virtual_feature_weights(self):
```
glinscott, line 41: Comment here about why we zero out the weights

model.py
```diff
@@ -20,31 +19,43 @@ class NNUE(pl.LightningModule):
...
+      return
+
+    # TODO: Implement this for more complicated conversions.
+    #       Currently we support only a single feature block.
+    if len(self.feature_set.features) > 1:
+      raise Exception('Cannot change feature set from {} to {}.'.format(self.feature_set.name, new_feature_set.name))
+
+    old_feature_block = self.feature_set.features[0]
+    new_feature_block = new_feature_set.features[0]
+    if old_feature_block.name == next(iter(new_feature_block.factors)):
```
glinscott, line 76: Hm..  I can't quite follow the logic here.  Can we add a bit of description about what is happening here?

glinscott:
Thanks!  Sorry for the delay, a fair bit of code to cover.  The structure looks good.  Perhaps can you create a features.md doc with the comment from your pull request talking about the architecture?  And also update the readme with the new flags that are needed?

I think the thing this really exposes is the trainer badly needs some tests :).  It's just too hard to reason about all the expected paths otherwise.  I have a branch where I started on that, and will do some more work on this today.

Has this been tested by training a network?  If so, LGTM after comments addressed.

Sopel97:
I've addressed your points and added some extra comments in other places where I saw necessary.

I have only briefly tested that this works (read "doesn't crash") and the loss goes down reasonably whether training from scratch or when resuming from .pt.

You make a good point about tests. The stuff around features at least is fairly testable, as it's easy to create mockup feature blocks and stuff. That's definitely something we should do in the near future.

docs/features.md
```diff
@@ -0,0 +1,15 @@
...
+
+The current semantics are as follows:
+
+1. When training a new net from scratch - `--features` specifies the feature set to use for learning. The feature transformer weights are initialized normally for the real features and zero initialized for the virtual features.
+2. When resuming training from a .pt model - `--features` specifies the feature set to use for learning. If the feature set specified doesn't match the feature set from the .pt model a conversion is attempted. Right now only a conversion of feature set with a single block from non-factorized to factorized is supported. The factorized block must have the non-factorized features as the first factor. The virtual feature weights are initialized to zero.
+3. When converting .ckpt to .nnue - `--features` specifies the features as stored in the .ckpt file. `load_from_checkpoint` doesn't correctly call the constructor with correct feature_set. I have not explored this and it may be improved, I don't know exactly how this works. The weights are coalesced when writing the .nnue file.
```
vondele: `load_from_checkpoint` doesn't correctly call the constructor with correct feature_set. I have not explored this and it may be improved, I don't know exactly how this works.

--> not clear to the average user what this means. Does it work, or not ?

Sopel97: Okay, I'll clarify. What it means is that the user must pass the correct feature set through `--features` because it can't be inferred from the .ckpt. If the features from `--features` and the saved model don't match it'll likely stack trace on some dimension mismatch.

vondele: OK, thanks.. yes, that is fine from my point of view.

vondele (approved):

glinscott (approved):

glinscott:
Thanks!  Feel free to merge anytime.

PR #34, opened 2020-12-12
Add support for (flip) HalfKA features

Sopel97:
This PR adds `HalfKA` and `HalfKA^` feature blocks.

The stockfish binary that can be used for playing (on nodchip because I couldn't be bothered with the official-stockfish): https://github.com/Sopel97/Stockfish/tree/flip_halfka

Note: This PR fixes the serialization and deserialization of the FT hash. Now the hash exposed by the feature class matches the hash in nodchip (previously it was ^512, which should be done later). This breaks .pt/.ckpt -> .nnue conversion of old models (but can be hacked by forcing model.feature_set in the code to be different, it just loads the old one from the checkpoint)

### DIFF
--- a/features.py
+++ b/features.py
@@ -10,8 +10,9 @@ function `get_feature_block_clss` at module scope that returns the list
 of feature block classes in that module.
 '''
 import halfkp
+import halfka
 
-_feature_modules = [halfkp]
+_feature_modules = [halfkp, halfka]
 
 _feature_blocks_by_name = dict()
 
--- a/halfka.py
+++ b/halfka.py
@@ -17,7 +17,7 @@ def halfka_idx(is_white_pov: bool, king_sq: int, sq: int, p: chess.Piece):
 
 class Features(FeatureBlock):
   def __init__(self):
-    super(Features, self).__init__('HalfKA', 0x5f134cb9, OrderedDict([('HalfKA', NUM_PLANES * NUM_SQ)]))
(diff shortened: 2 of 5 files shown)


Sopel97:
@vondele If you have some spare time for testing it would be nice to see how this fares against halfkp when trained to saturation.

vondele:
I'll be happily testing it, but right now I have no access to the testing box.

vondele:
sorry for the delay ... I forgot. As mentioned on discord, this architectures seems 20Elo stronger than HalfKP^ : 

I've trained a HalfKA^net on  my current best dataset, and it outperforms HalfKP^ by about 20 Elo. It is the strongest net I've trained so far (still a bit below master):
   # PLAYER                :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)
   1 master                :     0.0   ----  5691.5   10371    55     100
   2 run110/run0/halfka    :   -29.0    5.9  2385.5    5200    46     100
   3 run110/run1/halfka    :   -39.7    5.5  2294.0    5171    44     ---
(the two runs are independent optimizations, on the same data). Note also that this HalfKA^ is using the nodchip branch for playing, so just speed matters, and should add a few Elo.

vondele:
probably this could be merged ?

Sopel97:
I guess so. @glinscott could you give this a final review?

glinscott:
Sorry for the delay, LGTM!  Thanks for doing this!

PR #35, opened 2020-12-18
Compile the cpp data loader when importing the module.

Sopel97:
Cmake handles caching so is should be enough to run `compile_data_loader.bat` when the module `nnue_dataset` is being loaded. 

Now, while doing this I discovered that train.py requires the compiled data loader to be present regardless of whether the py or cc loader should be used. This should probably be looked into later, possibly import nnue_dataset conditionally?

### DIFF
--- a/nnue_dataset.py
+++ b/nnue_dataset.py
@@ -4,12 +4,34 @@ import torch
 import os
 import sys
 import glob
+import subprocess
 from torch.utils.data import Dataset
+
+commands = [
+    ['/bin/bash', 'compile_data_loader.bat'],
+    ['compile_data_loader.bat']
+]
+
+initialized = False
+for command in commands:
+    if initialized:
+        break
+
+    try:
+        subprocess.run(command, check=True)
+        initialized = True
+    except:
+        pass
+
+if not initialized:
+    print('Cannot compile data_loader shared library.')
+    sys.exit(1)
 
 local_dllpath = [n for n in glob.glob('./*training_data_loader.*') if n.endswith('.so') or n.endswith('.dll') or n.endswith('.dylib')]
 if not local_dllpath:
     print('Cannot find data_loader shared library.')
     sys.exit(1)
+
 dllpath = os.path.abspath(local_dllpath[0])
 dll = ctypes.cdll.LoadLibrary(dllpath)
 


Sopel97:
I see this PR still standing. Can you confirm it works correctly on systems other than windows? @vondele @glinscott

vondele:
for me, doesn't work. If I start without the .so present, it will stop, and the .bat script needs to be executed by bash. 

So, this would work on Linux for me, but won't work on windows, I assume:
```diff
diff --git a/nnue_dataset.py b/nnue_dataset.py
index 1d8eff6..51fc265 100644
--- a/nnue_dataset.py
+++ b/nnue_dataset.py
@@ -7,11 +7,11 @@ import glob
 import subprocess
 from torch.utils.data import Dataset
 
+subprocess.run(['/bin/bash','compile_data_loader.bat'])
 local_dllpath = [n for n in glob.glob('./*training_data_loader.*') if n.endswith('.so') or n.endswith('.dll') or n.endswith('.dylib')]
 if not local_dllpath:
     print('Cannot find data_loader shared library.')
     sys.exit(1)
-subprocess.run(['compile_data_loader.bat'])
 dllpath = os.path.abspath(local_dllpath[0])
 dll = ctypes.cdll.LoadLibrary(dllpath)
```

Sopel97:
Okay. I made a slightly uglier solution that should at least work.

vondele:
but  why is the compilation done after checking for the existing of the shared library? I think you want to do the check after you've tried to compile it ? Or do you want to force the user to compile the first time by hand?

Sopel97:
good point I completely missed that.

vondele:
BTW, I just now realize a potential issue with this patch. I tend to start 4 runs in the same directory at the same time (by script). This recompilation could be racy (multiple compiles going on at the same time, binaries starting while things are written), unless cmake does some magic to prevent that.

Sopel97:
hmm. indeed might be an issue. If the binary is up to date it's fine. If it's not the cmake call might fail, but shouldn't corrupt anything. Could probably be solved by a system wide mutex but not sure how doable this is.

Sopel97:
Since the data loader is now required this is not that big of an issue since it doesn't result in accidental bad performance. I'll close this.

PR #39, opened 2020-12-31
Add visualizer.

ddobbelaere:
This is a cleanup of the script in https://github.com/official-stockfish/Stockfish/issues/3274, extended to work on pytorch models (such that also checkpoints can be loaded).

At the moment, abs(input weights) is plotted (in the same style as the base script, but now normalized between 0 and 0.5).

Note that this is a "minimal working example". Please suggest if there is any need/wish to extend it with CLI arguments to e.g.

- Save plots directly to images (non-interactive mode for headless/automated use case).
- Control some parameters (saturation threshold that is now arbitrarily at 0.5; provide the option to plot non-abs values).
- Other useful plots?
- ...

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,14 @@ Import an existing SF NNUE network to the pytorch network format.
 python serialize.py nn.nnue converted.pt
 ```
 
+# Visualize a network
+
+Visualize a network from either a checkpoint (`.ckpt`), a serialized model (`.pt`)
+or a SF NNUE file (`.nnue`).
+```
+python visualize.py nn.nnue
+```
+
 # Logging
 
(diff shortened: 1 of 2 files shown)


Sopel97:
Since the visualization method depends on the feature set it would be better if we did not assume `HalfKP`. Perhaps the visualizer should be chosen based on feature set string, `visualizer_clss = { 'HalfKP' : NNUEVisualizer }`. For now we could even settle on simple `assert args.features == 'HalfKP'` and make a composable system like for the feature sets.

glinscott:
Thank you, this is great!  Will try out now :)

PR #41, opened 2021-01-02
Some visualizer improvements.

ddobbelaere:
Some visualizer improvements:

- Address https://github.com/glinscott/nnue-pytorch/pull/39#issuecomment-753008644
- Support configurable colormap range vmin/vmax (abs values are plotted if vmin >= 0). Default values are -0.5/0.5.
- `--save-dir` option to directly store plots in a given directory
- `--dont-show` option to not show the plots (for e.g. non-interactive use)
- Minor visual/layout improvements (slightly smaller colorbar, other colormap, tighter layout).

Suggestions/comments welcome!

### DIFF
--- a/visualize.py
+++ b/visualize.py
@@ -23,7 +23,7 @@ class NNUEVisualizer():
 
         return weight_coalesced
 
-    def plot_feature_transformer_weights(self, basename, vmin=0, vmax=0.5):
+    def plot_input_weights(self, basename, vmin=0, vmax=0.5, save_fig=False):
         # Coalesce weights and transform them to Numpy domain.
         weights = self.coalesce_ft_weights(self.model, self.model.input)
         weights = weights.transpose(0, 1).flatten().numpy()
@@ -40,6 +40,8 @@ class NNUEVisualizer():
         totaldim = totalx*totaly
 
         # Generate image.
+        print("Generating input weights plot...", end="", flush=True)
+
         # [Thanks to https://github.com/hxim/Stockfish-Evaluation-Guide
         # upon which the following code is based.]
         img = np.zeros(totaldim)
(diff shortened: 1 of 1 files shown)


ddobbelaere:
There are new insights to be gained by differentiating between positive/negative input weights (that's why I opted for default vmin/vmax values of -0.5/0.5).

Master/latest (best) vdv on fishtest:

image
image

Note that the weights of most "internal" features (of which there are 256) have opposite signs depending on us (left side) or theirs (right side) piece positions.

visualize.py
```diff
@@ -66,30 +68,57 @@ def plot_feature_transformer(self, title):
...
-        plt.title(title)
+        plt.xticks(ticks=widthx*np.arange(1, numx) - 0.5)
+        plt.yticks(ticks=widthy*np.arange(1, numy) - 0.5)
+        plt.axis('off')
+        plt.title(title_template.format(BASENAME=basename))
+        plt.tight_layout()
+
+        # Save figure.
+        if save_fig:
+            destname = "input-weights_{}.jpg".format(basename)
```
vondele: can we save in png format? lossless is nice for these plots.

I think it would also be nice to have the image in the same directory as the nets, probably also for the title used, in might be useful to have more of the path, otherwise all images will be named `last.nnue` in my case, while I will use `run73/run0/last.nnue` as argument.

ddobbelaere: Thanks for feedback, will try to address it soon.

Just a note: the resolution of the image at the moment is insufficient to losslessly represent the whole input weights matrix (which is 4096x2432). Apart from png, do we also want a "full resolution" option?

vondele: OK, didn't realize that it would result in a >20MB image, unless it is very easy, full resolution might be a bit overkill.

vondele:
BTW, the visualizer has already been quite useful IMO. Most of the discussion on the results is in discord. Obviously you can join there if you wish (https://discord.gg/K3rTpDzzuV)

ddobbelaere:
@vondele Thanks, I'm glad it helps. I really enjoy the research and the collaboration and I also try to keep myself up-to-date by following your excellent "pytorch training history" sheet.

As for discord, I currently lack the time, but also I know from experience I personnally tend to lose myself a bit and get too entangled in it. Been there, done that :)

glinscott:
This is wonderful, thank you so much!

visualize.py
ddobbelaere: Yeah, as it is currently not "lossless" anyway, I wanted to save some disk space (jpg is around 1MB, png around 2MB). So I left it at jpg...

I have addressed all your other comments with the options `--save-dir` (replaces `save-fig`), `--dont-show` and `--net-name`.

Sopel97:
Looks good.

PR #43, opened 2021-01-05
New params

vondele:
current set of params that leads to the best net from scratch so far

https://tests.stockfishchess.org/tests/view/5ff40fc86019e097de3eee8f
(run 79).

### DIFF
--- a/model.py
+++ b/model.py
@@ -97,10 +97,15 @@ class NNUE(pl.LightningModule):
   def step_(self, batch, batch_idx, loss_type):
     us, them, white, black, outcome, score = batch
 
-    q = self(us, them, white, black)
+    # 600 is the kPonanzaConstant scaling factor needed to convert the training net output to a score.
+    # This needs to match the value used in the serializer
+    nnue2score = 600
+    scaling = 361
+
+    q = self(us, them, white, black) * nnue2score / scaling
     t = outcome
-    # Divide score by 600.0 to match the expected NNUE scaling factor
-    p = (score / 600.0).sigmoid()
+    p = (score / scaling).sigmoid()
+
     epsilon = 1e-12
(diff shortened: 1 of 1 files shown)


model.py
```diff
@@ -97,10 +97,13 @@ def forward(self, us, them, w_in, b_in):
   def step_(self, batch, batch_idx, loss_type):
     us, them, white, black, outcome, score = batch
 
-    q = self(us, them, white, black)
-    t = outcome
-    # Divide score by 600.0 to match the expected NNUE scaling factor
-    p = (score / 600.0).sigmoid()
+    # 600.0 to match the expected NNUE scaling factor
+    nnue2score = 600
```
Sopel97, line 102: could we perhaps rename this, or at least document, that this is `kPonanzaConstant` from the nodchip trainer?

vondele: like:
```# 600.0 is the kPonanzaConstant scaling factor needed to convert the training net output to a score```
?

Sopel97: that and that it needs to match in stockfish

vondele: in stockfish, or in the serializer ?

Sopel97: Right, sorry, in the serializer. Stockfish doesn't need it obviously because it's baked into the weights.

Sopel97:
For completeness:
Master:
https://tests.stockfishchess.org/tests/view/5fee3c006019e097de3eea8d
https://tests.stockfishchess.org/tests/view/5feeebde6019e097de3eead0
https://tests.stockfishchess.org/tests/view/5fef930f6019e097de3eeb84
https://tests.stockfishchess.org/tests/view/5ff0b6f76019e097de3eec52

PR:
https://tests.stockfishchess.org/tests/view/5ff40fc86019e097de3eee8f
https://tests.stockfishchess.org/tests/view/5ff4ef836019e097de3eef3d

PR #44, opened 2021-01-11
Improved cmaps

vondele:
retain viridis for 'one-sided' visualization, use a diverging one for 2-side vis.

A two-sided one now looks like:

image

BTW, the fact that these weights look so symmetric with opposite sign, I'm wondering if we should initialize them similarly, i.e. as we do now, copy and invert those random numbers on the matching patches. Maybe there are other ways we could exploit this (like compress nets or so).

### DIFF
--- a/visualize.py
+++ b/visualize.py
@@ -78,8 +78,9 @@ class NNUEVisualizer():
 
         # Plot image.
         plt.figure(figsize=(16, 9))
+        cmap = 'coolwarm' if vmin < 0 else 'viridis'
         plt.matshow(img.reshape((totaldim//totalx, totalx)),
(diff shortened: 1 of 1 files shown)


Sopel97:
What's the entropy of all the differences? This is gonna be hard to use for compression in a general case, but if we find a simple way to define the relationship between such similar features then why not

ddobbelaere:
As for initialization, it can be worth a try. Note that this rule is not general though (there are quite a few spots where the symmetry isn't present).

As for compression, also good idea. Maybe we can just try to flip signs of black piece weights and see how the net compresses, without the hassle of differences that might not fit into int16? Most modern compression tools like xz have look-ahead buffers on the order of tens of megabytes for highest compression levels.

Issue #47, opened 2021-01-21
Idea for reducing quantization error


ddobbelaere:
This evening I had some fun improving the visualizer. I'll try to create a PR if I find the time for it.

One of the things I investigated are the weights of the fully-connected layers after the input layer. Also, as all hidden neurons are "arbitrarily ordered" (a random permutation leads to an equivalent net), I ordered the input weights by their L1-norm (sum of absolute values).

Here are the plots for master net nn-62ef826d1a6d.nnue (an offspring from sergioviero's run) and nn-0f63c1539914.nnue (latest vdv net on fishtest).

![master](https://user-images.githubusercontent.com/38748663/105403771-7a49bf00-5c29-11eb-8411-676424118bf1.png)
![master_fc](https://user-images.githubusercontent.com/38748663/105403789-7fa70980-5c29-11eb-9f72-50ea034c0302.png)
![master_hist](https://user-images.githubusercontent.com/38748663/105403816-8766ae00-5c29-11eb-9a8d-739aa1c71c34.png)
![vdv](https://user-images.githubusercontent.com/38748663/105403776-7c138280-5c29-11eb-830c-fbe7080c013f.png)
![vdv_fc](https://user-images.githubusercontent.com/38748663/105403792-8170cd00-5c29-11eb-8f05-c39ec10e7839.png)
![vdv_hist](https://user-images.githubusercontent.com/38748663/105403820-8897db00-5c29-11eb-8dec-613a8bb5284c.png)

Here's the main "observation": the (quantized) FC1 weights are sharply peaked (even more so for vdv net than for master) around zero. This is not the case for FC2 weights (not shown here, but already "visually apparent" from other figure). I suspect that the quantization error leads to some ELO loss (how much is the question :)).

Would it be a good idea to rescale the input weights such that the FC1 weights become higher in magnitude, such that their quantization error decreases? Denote the output of neuron j at layer i as $x^{(i)}_j$, the bias term $b_j$ and its input weights $a_k$. With a non-linear activation function $sigma$, this leads to the familiar equation:

![image](https://user-images.githubusercontent.com/38748663/105405614-a108f500-5c2b-11eb-825d-70be9364dbdd.png)

If sigma is a clamped ReLU and q_j>0, it holds that

![image](https://user-images.githubusercontent.com/38748663/105406316-8be09600-5c2c-11eb-941a-b75d680b5ad2.png)

if and only if sigma "is not clamping" (to the max. saturation value).

[...]

PR #48, opened 2021-01-23
Add visualization for weights of final layers.

ddobbelaere:
This a cleanup version of of the code used in #47, with the following additions/improvements:
- Heatmap visualization of the weights in the final three layers.
- Histogram plots of the weights in the first and second fully-connected layers.
- Added argument `--order-input-neurons` to order the input neurons based on the L1-norm (sum of absolute values) of their connected weights. This norm seemed most "visually pleasing" (I also tested L2-norm and L_infinity-norm (max abs norm), but didn't like them).
- Changed default vmin/vmax for input weights to -1/+1, as this seems more logical and saner for most of the nets that are currently generated in the training runs.
- Add prefix for saved files (default = net name), can be overriden with `--save-prefix`.

Sample outputs:

input-weights
fc-weights
l1-weights-histogram
l2-weights-histogram

Suggestions are welcome. I am currently considering to maybe represent the L1 weights a bit differently, such that weights of the (own, opponent) input neurons are more close to each other (EDIT: done).

### DIFF
--- a/visualize.py
+++ b/visualize.py
@@ -252,7 +252,7 @@ def main():
     features.add_argparse_args(parser)
     args = parser.parse_args()
 
-    assert args.features == 'HalfKP' or args.features == 'HalfKP^'
+    assert args.features in ['HalfKP', 'HalfKP^']
     feature_set = features.get_feature_set_from_name(args.features)
 
     print("Visualizing {}".format(args.source))
--- a/visualize.py
+++ b/visualize.py
@@ -4,6 +4,7 @@ import model as M
 import numpy as np
 import torch
 import matplotlib.pyplot as plt
+from matplotlib.gridspec import GridSpec
 
 from serialize import NNUEReader
 
@@ -29,6 +30,7 @@ class NNUEVisualizer():
         weights = weights.transpose(0, 1).flatten().numpy()
 
         hd = 256  # Output feature dimension.
+        self.M = hd
         numx = 32  # Number of output features per row.
 
         self.ordered_input_neurons = np.arange(hd, dtype=int)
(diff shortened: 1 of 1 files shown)


vondele:
I have one question on the range of the weights. The historgrams suggest they go up to 2, in the integer range we have -127,128. Are you using the `kMaxWeight = 127.0 / kWeightScale # roughly 2.0` range ?

ddobbelaere:
@vondele For the histograms, I set the edges of the bins *between* "integers divided by 64", as I think `kWeightScale = kBiasScale / kActivationScale # = 64.0 for normal layers` holds exactly. Please tell me if this is an incorrect reasoning. The +/-2 histogram edges are only approximately valid and set automatically by matplotlib.

Btw, I have also updated the heatmap default vmin/vmax for FC layers to +/-2 in latest commit.

Sopel97:
For the feature transformer the weights can go up to around 258.0 in magnitude, but in practice they are much closer to 0. We observe min/max pre activations to be ~15 for well trained nets which suggests the weights average about 0.5.

For the affine transform layers the domain is defined by:
```
// Coefficient used for parameterization
        static constexpr LearnFloatType kActivationScale =
            std::numeric_limits<std::int8_t>::max();

        static constexpr LearnFloatType kBiasScale = kIsOutputLayer ?
            (kPonanzaConstant * FV_SCALE) :
            ((1 << kWeightScaleBits) * kActivationScale);

        static constexpr LearnFloatType kWeightScale = kBiasScale / kActivationScale;

        // Upper limit of absolute value of weight used to prevent overflow when parameterizing integers
        static constexpr LearnFloatType kMaxWeightMagnitude =
            std::numeric_limits<typename LayerType::WeightType>::max() / kWeightScale;
```

So for the hidden layers it's `+-127/64 ~= +-1.98`

and for the output layer it's `+-127/(600*16/127) ~= +-1.68`

Sopel97 (approved):

ddobbelaere:
Now also HalfKP^ is allowed (as the input weights get coalesced before plotting anyway), so that also checkpoints can be visualized (this is not possible in current master).

PR #52, opened 2021-01-30
Bugfix in ordering of L1 weights + also order input biases.

ddobbelaere:
Too weak, too slow :)

I noticed a bug in the visualizer if `--order-input-neurons` is set. Now the L1 weights and input biases are also properly reordered.

Following up on the input biases optimization discussion, we might also consider optimizing a subset (e.g. 32) of the 256 input biases, assuming neurons with the highest weights norm are "more important". A reordered master net (ordered on L1 norm) looks like this:

reordered nn-62ef826d1a6d nnue_input-weights
reordered nn-62ef826d1a6d nnue_biases

### DIFF
--- a/visualize.py
+++ b/visualize.py
@@ -55,7 +55,7 @@ class NNUEVisualizer():
         self.M = hd
         numx = 32  # Number of output features per row.
 
-        self.ordered_input_neurons = np.arange(hd, dtype=int)
+        inv_ordered_input_neurons = np.arange(hd, dtype=int)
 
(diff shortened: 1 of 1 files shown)


vondele:
I had a script to compute the sensitivity of the output wrt to a net weight/bias params. 

Basically, compute `(1/N) sum_i (d output_eval_i / d param_k)^2` over a set of large fens... I used finite difference for the derivative, in principle could be analytic now. I think that's the better measure of 'most important'.

PR #54, opened 2021-01-30
Fix script if --order-input-neurons not set.

ddobbelaere:
Sorry, was too hasty in previous PR and didn't check the case if `--order-input-neurons` is not set.
It just crashes right now in that case:

`AttributeError: 'NNUEVisualizer' object has no attribute 'ordered_input_neurons'`


Sopel97:
is the removal of some parameters desired?

ddobbelaere:
@Sopel97 yes, they are unused after the switch to the `--label` argument, which is now used to override both the plot label and save prefix.

PR #56, opened 2021-02-01
Fix virtual feature initialization

sergiovieri:
The weight shape of torch.nn.Linear is (out, in) ([source](https://pytorch.org/docs/stable/generated/torch.nn.Linear.html#torch.nn.Linear)).

In the case of HalfKP^, `print(a, b, weights.shape)` produces `41024 41728 torch.Size([256, 41728])`.


ddobbelaere:
Good catch! And also here I guess: https://github.com/glinscott/nnue-pytorch/blob/6cf71b15879509c367b086a97bbab29a78e0da9f/qmodel.py#L54

PR #58, opened 2021-02-02
Python script to run games to find best net

vondele:


vondele:
I am making a pull requests for comments / testing. Not sure if we want this in this repo, but works nicely for me.

vondele:
Typical Elo evolution, with error bars as obtained by running the script during a training run. 
image
It will produce also a list of best nets:
```
Found 104 nn-epoch*.nnue files
Best nets so far:
   run96/run1/nn-epoch226.nnue : -18.6 +- 2.5
   run96/run1/nn-epoch353.nnue : -18.9 +- 4.0
   run96/run1/nn-epoch290.nnue : -21.1 +- 3.8
Measuring nets:
   run96/run1/nn-epoch353.nnue : -18.9 +- 4.0
   run96/run1/nn-epoch226.nnue : -18.6 +- 2.5
   run96/run1/nn-epoch298.nnue : -21.5 +- 4.3
Running match with c-chess-cli ... run96/out.pgn
Running ordo ranking ... run96/ordo.out

```

glinscott:
Very nice!

run_games.py
```diff
@@ -0,0 +1,232 @@
...
+                ckpts.append(os.path.join(path, filename))
+
+    # lets move the .nnue files a bit up in the tree, and get rid of the = sign.
+    # run96/run0/default/version_0/checkpoints/epoch=3.ckpt -> run96/run0/nn-epoch3.nnue
+    for ckpt in ckpts:
+        nnue_file_name = re.sub("default/version_[0-9]+/checkpoints/", "", ckpt)
+        nnue_file_name = re.sub("epoch=", "nn-epoch", nnue_file_name)
+        nnue_file_name = re.sub(".ckpt", ".nnue", nnue_file_name)
+        if not os.path.exists(nnue_file_name):
+            command = "python3 serialize.py {} {} ".format(ckpt, nnue_file_name)
```
Sopel97, line 25: maybe could use `sys.executable` to get the location of the python.exe?

vondele: https://github.com/glinscott/nnue-pytorch/pull/59

tryingsomestuff:
I saw ordo stall once wile running the tool. Took 100% cpu for 35minutes ...

So maybe using something like this instead of just os.system might be safer:

```
class Command(object):
    def __init__(self, cmd):
        self.cmd = cmd
        self.process = None

    def run(self, timeout):
        def target():
            self.process = subprocess.Popen(self.cmd, shell=True, preexec_fn=os.setsid)
            self.process.communicate()

        thread = threading.Thread(target=target)
        thread.start()

        thread.join(timeout)
        if thread.is_alive():
            os.killpg(self.process.pid, signal.SIGTERM)
            thread.join()
        return self.process.returncode
```

Issue #53, opened 2021-01-30
Large input weights fluctuations for different king positions


ddobbelaere:
I'm trying to get some more grip onto the input layer and decided to change the order of piece and king indices (along with undoing the king index flipping, in order to not let my brain explode in interpreting the plots). Here's the diff w.r.t. master:

```
--- a/visualize.py
+++ b/visualize.py

-                inpos = [(7-kipos[0])+pipos[0]*8,
-                         kipos[1]+(7-pipos[1])*8]
-                d = - 8 if piece < 2 else 48 + (piece // 2 - 1) * 64
+                inpos = [8*kipos[0]+pipos[0],
+                         8*(7-kipos[1])+(7-pipos[1])]
+                d = -2*(7-kipos[1]) - 1 if piece < 2 else 48 + \
+                    (piece // 2 - 1) * 64
```

The plots become:

![reordered nn-62ef826d1a6d nnue_input-weights](https://user-images.githubusercontent.com/38748663/106363185-5d974080-6327-11eb-8ae5-fde6a12bd7ac.jpg)
![reordered epoch=427 ckpt_input-weights](https://user-images.githubusercontent.com/38748663/106363187-5f610400-6327-11eb-978c-395848a5c0c0.jpg)

Some strange things/observations:
- ~There is something special going on with king on first rank for the queen not on first rank weights, both in master and in vondele net. Example hidden input neuron (there are a lot more of them):~ BUG in code, forgot to update offset...

- Look at this weird pattern in vondele net:
![image](https://user-images.githubusercontent.com/38748663/106363184-5839f600-6327-11eb-9770-973ca71bf899.png)
The only explanation I can give for it is that the hidden input neuron serves the purpose of detecting two "hidden features" at once. If the king is on the back rank (or on a6, a7, d7, h6 or h7), something completely different is happening/detected. Either that, or some gap in training data?

vondele:
this reminds me of your correlation plots, which has this strange pattern for my nets https://github.com/glinscott/nnue-pytorch/issues/45. 

The other thing is, in pytorch we are skipping all fens where we are in check (or the bestmove is a capture).. could these features correspond to e.g. being in check?

ddobbelaere:
> The other thing is, in pytorch we are skipping all fens where we are in check (or the bestmove is a capture).. could these features correspond to e.g. being in check?

Hmm, not really I think. The "in check" weights are the smaller patterns in each 8x8 subsquare that represent one king position.

[...]

PR #61, opened 2021-02-07
Add king-centric input weights order to visualizer.

ddobbelaere:
Several visualizer changes:
- Added king-centric input weights order via new argument `--input-weights-order`. This allows us to investigate some issues better, e.g. #53.
    - `piece-centric-flipped-king` (default), as used currently, same as https://github.com/hxim/Stockfish-Evaluation-Guide
    - `king-centric` (one block matrix per king position, for each piece position).
- Renamed `--order-input-neurons` to `--sort-input-neurons`
- Fix label for histogram in combination with `--input-weights-vmin >= 0` (then also abs(input weights) is shown).

Default:
default_nn-factorizer-off-epoch83 nnue_input-weights
 
With `--sort-input-neurons --input-weights-order=king-centric`:
nn-factorizer-off-epoch83 nnue_input-weights

### DIFF
--- a/visualize.py
+++ b/visualize.py
@@ -55,21 +55,21 @@ class NNUEVisualizer():
         self.M = hd
         numx = 32  # Number of output features per row.
 
-        inv_ordered_input_neurons = np.arange(hd, dtype=int)
+        inv_sorted_input_neurons = np.arange(hd, dtype=int)
 
-        if self.args.order_input_neurons:
-            # Order input neuron by the L1-norm of their associated weights.
+        if self.args.sort_input_neurons:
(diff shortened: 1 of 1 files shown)


vondele:
I'd have a wish for this PR... can we have a some kind of axis labels for this... I always have a hard time to figure out what we see. Bonus points for a 'mouse-over' window that kind of shows where we point at ...

ddobbelaere:
@vondele Nice suggestions. I like the hover-over info (for all graphs maybe), but that'll take some time to figure out I think. Maybe this is best handled in another PR?

vondele:
sure..

PR #63, opened 2021-02-07
Add realtime results to run_games.py

glinscott:
Also, ctrl+c now works to stop run_games.py

Gives output like this:
```
Score of master vs logs/nn-epoch147.nnue: 13 - 7 - 29  [0.561] 49
Score of master vs logs/nn-epoch145.nnue: 11 - 0 - 18  [0.690] 29
```

### DIFF
--- a/run_games.py
+++ b/run_games.py
@@ -1,5 +1,6 @@
 import re
 import os
+import subprocess
 import sys
 import time
 import argparse
@@ -64,7 +65,6 @@ def parse_ordo(root_dir, nnues):
 def run_match(best, root_dir, c_chess_exe, concurrency, book_file_name, stockfish_exe):
     """ Run a match using c-chess-cli adding pgns to a file to be analysed with ordo """
     pgn_file_name = os.path.join(root_dir, "out.pgn")
-    c_chess_out_file_name = os.path.join(root_dir, "c_chess.out")
     command = "{} -each tc=4+0.04 option.Hash=8 option.Threads=1 -gauntlet -games 200 -rounds 1 -concurrency {}".format(
         c_chess_exe, concurrency
     )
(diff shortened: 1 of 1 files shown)


run_games.py
```diff
@@ -79,13 +79,26 @@ def run_match(best, root_dir, c_chess_exe, concurrency, book_file_name, stockfis
...
-        pgn_file_name, c_chess_out_file_name
+    command = command + " -pgn {} 0 2>&1".format(
+        pgn_file_name
     )
 
     print("Running match with c-chess-cli ... {}".format(pgn_file_name), flush=True)
-    ret = os.system(command)
-    if ret != 0:
+    c_chess_out = open(os.path.join(root_dir, "c_chess.out"), 'w')
+    process = subprocess.Popen("stdbuf -o0 " + command, stdout=subprocess.PIPE, shell=True)
```
vondele, line 88: this might make it difficult to run the script on windows? @Sopel97 might want to comment

Sopel97: maybe this can be achieved by passing `bufsize=0` to `Popen`? https://docs.python.org/3/library/subprocess.html#subprocess.Popen

```
bufsize will be supplied as the corresponding argument to the open() function when creating the stdin/stdout/stderr pipe file objects:
    0 means unbuffered (read and write are one system call and can return short)
    1 means line buffered (only usable if universal_newlines=True i.e., in a text mode)
    any other positive value means use a buffer of approximately that size
    negative bufsize (the default) means the system default of io.DEFAULT_BUFFER_SIZE will be used.
```

glinscott: Unfortunately not, I tried that - I think it's because of shell=True, so it ends up setting the buffer size for the shell process.  I could try getting rid of shell=True so there isn't a wrapping shell process.

vondele: actually, I realize c-chess-cli doesn't even support windows, so maybe this is a non-issue?

PR #64, opened 2021-02-08
[RFC] New storage format for NNUE nets.

Sopel97:
This is a reference implementation for a possible new storage format for NNUE nets. The file nnuebuf.py contains the rough outline of the schema at the top. The purpose of this specification is to be as simple as possible while supporting all reasonable cases, allowing parsing (round trip) with limited knowledge, and being extensible.

Right now nnuebuf.py also works as a tool for conversion from current format to the new one, and running round trip tests.
This version keeps some verbose printing for debugging purposes.

`python nnuebuf.py infile.nnue outfile [-c] [--test]`
`-c` - compress the feature transformer weights (gets master net to ~7.8MB)
`--test` - do round trip.

Some stuff that is trivially supported with this:
- storing nets in a factorized format - just standardize HalfKP^ and stuff
- storing nets before quantization - just use DataType.float32
- ignoring unknown feature blocks - field size is encoded
- ignoring unknown layers - field size is encoded
- storing multiple nets
- having multiple sub nets after the feature transformer
- encoding and validation of feature transformer semantics

I would like to get some general feedback on this. Do you see this being a thing? Anything missing? Anything not clear? Anything could be improved code-wise? Anything ambiguous? Anything unspecified? Any ideas what to do with stuff marked "TODO"?

### DIFF
--- a/nnuebuf.py
+++ b/nnuebuf.py
@@ -1,0 +1,1689 @@
+import numpy as np
+import enum
+import struct
+import io
+import math
+
+'''
+General schema:
+
+int64
+    payload : byte[8]
+        little endian
+
+uint64
+    payload : byte[8]
+        little endian
+
+int32
+    payload : byte[4]
+        little endian
+
+uint32
+    payload : byte[4]
+        little endian
+
+uint16
+    payload : byte[2]
+        little endian
+
+int16
+    payload : byte[2]
+        little endian
+
+uint8
+    payload : byte
+
+int8
+    payload : byte
+
+float16
+    payload : byte[2]
+        IEEE754
+
+float32
+    payload : byte[4]
+        IEEE754
+
+float64
+    payload : byte[8]
+        IEEE754
+
+bool
+    payload : byte
+        0 - false
+        1 - true
+        any other value is not allowed
+
+ShortString
+    size : uint8
+    payload : size bytes
+        utf-8
+
+LongString
+    size : uint16
+    payload : size bytes
+        utf-8
+
+DataType
+    enum : uint8
+        0 - uint8
+        1 - int8
+        2 - uint16
+        3 - int16
+        4 - uint32
+        5 - int32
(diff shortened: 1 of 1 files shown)


vondele:
so, I think this is indeed a useful direction. Not sure I can judge the code, but some of the features would be useful.

One potential field missing, for consideration, would we want a shortString License ?

Sopel97:
License is a good idea. Do you think 255 characters just for the name + link is okay? If not we can allow up to 65535

vondele:
probably we could call the field 
SPDX-License-Identifier: 
and suggest that the field is from
https://spdx.dev/licenses/
which includes the CC0-1.0 and other licenses and allows for UNLICENSED or SEE LICENSE IN <filename>

Sopel97:
Maybe add some specification for the routing algorithm done in the net to choose a layer stack to use for a position. This could be anything, from a complex description to a layer-like router. Just for completeness.

Sopel97:
Since very different quantization schemes might be feasible (as Gary is trying some stuff) it might be needed to have a field specifying the quantization used. Probably will have to specify quantization 'style' for every layer, with custom parameters for each 'style'. If there is some quantization specification it should be 'flexible', as layer type names are, so be a short string.

Sopel97:
Now I'm experimenting with some stuff where the feature transformer collects an additive evaluation term that is used directly in the output. This is essentially like PSQT but tied to features. I could see this becoming a thing and so this format would need to support it in some way that's not terribly complicated. The least invasive way I see right now would be to allow for additional rows (256 for our net, 256+1 for our net + psqt) and these additional rows would have separate type information for each one of them. Compression would work the same for them as for the normal weights of the feature transformer.

Sopel97:
I'll close this as it is too complex, slow, and not quite as easily extensible to new ideas as previously assumed. Effort to improve on this is underway.

PR #67, opened 2021-02-14
Support visualization of arbitrary number of input neurons.

ddobbelaere:
Arbitrary L2 and L3 were already supported, now also support arbitrary L1.

For the input weights plot, the ratio of the number of input neurons per row/col is chosen such that is as close to 4 (the current value for 256) as possible. This leads to the following aspect ratios for some likely candidates:

L1 | ratio
---- | ----
128 | 2
192 | 3
256 | 4
384 | 2.66...
512 | 2

Tested with a random 512x16x16 net (just change L1, L2, L3 in model.py):

some_net nnue_input-weights

### DIFF
--- a/visualize.py
+++ b/visualize.py
@@ -17,10 +17,10 @@ class NNUEVisualizer():
         self.args = args
 
         import matplotlib as mpl
-        dpi = 100
+        self.dpi = 100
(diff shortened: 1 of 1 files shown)


vondele:
cool, we might have some interesting 512x16x16x1 nets to visualize eventually.

PR #68, opened 2021-02-14
Fix NNUEReader if SIMD weights padding is present.

ddobbelaere:
Complements 36de78f8729fe48f97fe81756ce65e2e5700e4c6
Amongst other things, fixes bogus output bias term in visualization if padding in L3 is present.

Now everything is in place to correctly visualize e.g. 512x16x16 nets :)


glinscott:
Thanks for fixing these issues!

PR #69, opened 2021-02-14
Add changes for different net arch

vondele:
512x16x16x1

To be merged only after some fishtest results are added.


vondele:
https://tests.stockfishchess.org/tests/view/60293ef47f517a561bc49984 good but not better than master. Closing for now.

PR #70, opened 2021-02-15
Add support for different binaries to run_games.py

glinscott:
Useful when testing out new architectures.


Sopel97:
maybe `stockfish_base`, `stockfish_test` to be consistent with fishtest?

vondele:
good idea, I agree base and test is a better naming.

glinscott:
Sounds good, updated.

PR #78, opened 2021-03-13
Add a script for plotting loss and ordo elos.

Sopel97:
Usage: `python do_plots.py out_file.png root_dir0 [root_dir1] [root_dir2] ...`

`root_dir`s are searched for:
1. tfevents file
2. ordo.out

`val_loss` and `train_loss` is extracted from tfevents file. If more than one value per step was found in the given `root_dir` (for example if there are multiple ftevents files) the best value is taken for plotting.

if `ordo.out` exists (otherwise it's not used) it is parsed and the best result for each epoch is used for plotting the elo.

example output:
plot_10_11

### DIFF
--- a/do_plots.py
+++ b/do_plots.py
@@ -1,0 +1,190 @@
+import numpy as np
+
+from tensorboard.backend.event_processing.event_accumulator import EventAccumulator
+
+import matplotlib.pyplot as plt
+import argparse
+import re
+import sys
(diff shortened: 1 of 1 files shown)


Sopel97:
better argument parsing is to be done.

glinscott:
Nice, merge away :).

PR #81, opened 2021-03-15
Add a script for deleting worst nets within a directory based on ordo.

Sopel97:
Directly from help:
```
Usage: python delete_bad_nets.py root_dir [num_best_to_keep]
root_dir - the directory to "cleanup"
num_best_to_keep - the number of best nets to keep. Default: 16

It expects to find ordo.out somewhere within root_dir.
If the ordo.out is not found nothing is deleted.
It uses the ratings from the ordo file to determine which nets are best.
The engine names must contain the network name in the
following format: "nn-epoch[0-9]*\.nnue". The network file
can be specified with a parent directory (for example
"run_0/nn-epoch100.nnue"), in which case the .ckpt file corresponding
to this .nnue file will only be searched for in the parent ("run_0") directory.
The .ckpt files must contain "epoch=([0-9]*).*\.ckpt".
Both ckpt and nnue files are deleted. Only nets listed in the ordo
file can be deleted. Other nets are always kept.
The .nnue and .ckpt files are matched by epoch.

The directory layout can be for example:
- root_dir
  - run_0
    - a/b/c/d.ckpt
    - *.nnue
  - run_1
    - a/b/c/d.ckpt
    - *.nnue
  - ordo.out
    (in this case ony lines with engine name matching
     "run_[01]/nn-epoch[0-9]*\.nnue" will be used.)
```

I think the assumptions on the directory structure and naming are reasonable enough for this to be useful to others. I'll sometime soon probably also put a set of .sh scripts I'm using that match this.

### DIFF
--- a/delete_bad_nets.py
+++ b/delete_bad_nets.py
@@ -1,0 +1,151 @@
+import sys
+import re
+import os
+import itertools
+
+def parse_ordo(ordo_filename):
+    ordo_scores = []
+
+    with open(ordo_filename, 'r') as ordo_file:
+        lines = ordo_file.readlines()
+        for line in lines:
+            if 'nn-epoch' in line:
+                fields = line.split()
+                net = fields[1]
+                rating = float(fields[3])
+                error = float(fields[4])
+                ordo_scores.append((net, rating, error))
+
+    return ordo_scores
+
+def find_ckpt_files(root_dir):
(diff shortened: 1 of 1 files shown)


glinscott:
Oh, apologies for the delay, LGTM!

PR #85, opened 2021-04-01
A new setup.

Sopel97:
I've come up to a setup that to my knowledge produced the best 256x2-32-32-1 net and the best 512x2-16-32-1 net yet.

256x2-32-32-1: https://tests.stockfishchess.org/tests/view/60659d0b2b2df919fd5f028d (using hybrid, tested directly against master; so the usual setup for STC tests). Same test with the newly possible NNUE-PSQT hybrid https://tests.stockfishchess.org/tests/view/60659dab2b2df919fd5f028f
512x2-16-32-1: https://tests.stockfishchess.org/tests/view/6065a5bd2b2df919fd5f02a0 (using a psqt hybrid, tested directly against master; usual STC setup but hybrid approach is different)

The changes are the following:
1. Initialization of biases is changed. Hidden neuron biases are initialized to half of the activation range. Output bias to 0.
2. Change architecture from HalfKP to HalfKA (important to have features for kings, and flip instead of rotate) (still uses the 641 features per plane and wastes one plane (so it could be made ~8% smaller). This can be improved in the future, but would need a new name, for example HalfKAv2. This is for another PR.
3. Weights are clipped after each step to the range of quantization. (I changed the ranger.py file. Does it need any changes to the header due to these changes?)
4. An additional value is added for each feature, with the idea for it to represent the PSQT value. Value of this feature is directly forwarded to the output. With how the NNUE works it is necessary for this value to be evaluated for both perspectives and the average be chosen.
5. Change scaling of the eval->perf% sigmoid for the input data from 361 to 410 (which matches closer the empirical value for depth 9 data, which is the most popular right now. This still should be adjusted for other data, but the effect seems to be small.)

The stockfish branch which can use these nets can be found here: https://github.com/Sopel97/Stockfish/tree/new_setup

More data and history for these changes can be found in https://docs.google.com/document/d/1gTlrr02qSNKiXNZ_SuO4-RjK4MXBiFlLE6jvNqqMkAY/edit?usp=sharing.

### DIFF
--- a/feature_set.py
+++ b/feature_set.py
@@ -102,3 +102,9 @@ class FeatureSet:
             real_offset += feature.num_real_features
             offset += feature.num_features
         return indices
+
+    def get_initial_psqt_features(self):
+        init = []
+        for feature in self.features:
+            init += feature.get_initial_psqt_features()
+        return init
--- a/features.py
+++ b/features.py
@@ -40,7 +40,7 @@ def get_available_feature_blocks_names():
     return list(iter(_feature_blocks_by_name))
 
 def add_argparse_args(parser):
-    _default_feature_set_name = 'HalfKP^'
+    _default_feature_set_name = 'HalfKA^'
     parser.add_argument("--features", dest='features', default=_default_feature_set_name, help="The feature set to use. Can be a union of feature blocks (for example P+HalfKP). \"^\" denotes a factorized block. Currently available feature blocks are: " + ', '.join(get_available_feature_blocks_names()))
 
 def _init():
--- a/halfka.py
+++ b/halfka.py
@@ -15,6 +15,28 @@ def halfka_idx(is_white_pov: bool, king_sq: int, sq: int, p: chess.Piece):
   p_idx = (p.piece_type - 1) * 2 + (p.color != is_white_pov)
   return 1 + orient(is_white_pov, sq) + p_idx * NUM_SQ + king_sq * NUM_PLANES
 
+def halfka_psqts():
+  # values copied from stockfish, in stockfish internal units
+  piece_values = {
+    chess.PAWN : 126,
+    chess.KNIGHT : 781,
+    chess.BISHOP : 825,
+    chess.ROOK : 1276,
+    chess.QUEEN : 2538
+  }
+
+  values = [0] * (NUM_PLANES * NUM_SQ)
+
+  for ksq in range(64):
+    for s in range(64):
+      for pt, val in piece_values.items():
+        idxw = halfka_idx(True, ksq, s, chess.Piece(pt, chess.WHITE))
+        idxb = halfka_idx(True, ksq, s, chess.Piece(pt, chess.BLACK))
+        values[idxw] = val
+        values[idxb] = -val
+
+  return values
+
 class Features(FeatureBlock):
   def __init__(self):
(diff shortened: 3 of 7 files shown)


vondele:
I'll give this PR a try locally and see what I get...

vondele:
First review comment: It would need an update of the Readme:

https://github.com/glinscott/nnue-pytorch#current-recommended-training-invocation

Sopel97:
The visualization command would need an update. I'm actually not sure it'll work out of the box, will have to check.

vondele:
So, I've done a number of tests, the two best ones were both about -10Elo from master on fishtest at 10+0.1s :
https://tests.stockfishchess.org/tests/view/6068d3312b2df919fd5f0441
https://tests.stockfishchess.org/tests/view/606bec06de8d5df05f51a077
some of the nets outperform master at VSTC.
The best nets of 8 different runs at short TC locally span about 20Elo:
```
   1 run203/run3/nn-epoch387.nnue    :     3.0    3.0    7564.0   15000    50      81
   2 run202/run0/nn-epoch233.nnue    :    -1.6    0.9   83178.5  167134    50      76
   4 run202/run2/nn-epoch181.nnue    :    -3.3    1.8   26451.5   53402    50      72
   5 run202/run3/nn-epoch183.nnue    :    -4.1    2.3   16305.0   33000    49      59
  11 run203/run2/nn-epoch155.nnue    :    -6.6    2.8    9223.0   18800    49      55
  23 run203/run1/nn-epoch255.nnue    :   -11.0    6.8    1937.0    4000    48      58
  28 run202/run1/nn-epoch139.nnue    :   -11.1    5.4    3196.0    6600    48      53
  39 run203/run0/nn-epoch281.nnue    :   -17.0   10.0     696.5    1464    48      53
```
That's better than the best net with the default setup so far, but it comes with the disadvantage that one need has a different architecture. Not so sure we should merge this right away.

vshcherbyna:
Tomasz, out of curiosity I give a try for this new version by doing training on my existing Igel data. I noticed that the training speed is roughly three times slower than on the master branch. Is this is expected? I double check everything, same data, same SSD, but training is 3 times slower still.

Sopel97:
I don't see any immediate reason why it would be slower. I have not measured any slowdown on GPUs

vshcherbyna:
_I don't see any immediate reason why it would be slower. I have not measured any slowdown on GPUs._

My current mainstream training is also on GPU, I did triple check, I see a clear slowdown between master and this branch. Perhaps this is related also to the cuda version inside the venv, I will double check the difference between the two.

Thanks for the link to profiler, I will check it out too.

Sopel97:
Are you perhaps running out of VRAM? This branch requires very close to 2GB

vshcherbyna:
My machine has 32 Gb of RAM and at the time of the training only 6 GB are used. 

I will do more testing over weekend to see if I can find the root cause of this.

vondele:
I assume the remark concerned the amount of GPU memory (so have 2GB of memory on the GPU available for training).

vshcherbyna:
Right, the GPU was idle before the training and it has 8GB of dedicated memory and 16 GB of shared. Typically when my machine is idle 1.3 GB of 8 GB of dedicated memory consumed. So should be plenty of memory available for the training.

P.S. I am doing this in latest Windows 10.

PR #86, opened 2021-04-02
A set of scripts for training.

Sopel97:
I took the scripts I use, made them configurable, and overall improved the usability. 

I place them in a new directory because they are meant to be used a set and don't depend on anything else from the repo (ofc it does use the trainer, but doesn't call anything from the repo directory directly and doesn't assume paths).

I'll be able to test them only tomorrow, as currently I'm running some experiments. I put it as a draft to get some general feedback.

There's a README.md in there, I'd like to know if everything is clear from it and the config.sh

### DIFF
--- a/scripts/1/README.md
+++ b/scripts/1/README.md
@@ -7,6 +7,7 @@ The basic principles are the following:
 - each experiment is self contained
     - contains the code of the trainer used
     - contains the code of the player used
+    - contains the scripts
 - this script directory is self contained
     - it doesn't require anything from outside to be under a specific path
 - only copied files are executed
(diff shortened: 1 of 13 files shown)


Sopel97:
I've tested and fixed the scripts.

Sopel97:
I've rebased on the current master and updated the plotting scripts

Sopel97:
I've revised how these are supposed to be used. Now ./setup_experiment must be used first, and it will copy all the scripts to the experiment_ID/scripts/ directory, from where the user is supposed to do further executions. This means the train/run_games infers the experiment it's in. Also, both base and test stockfish's are now copied in source form and compiled - this requires two directories with 2 (possibly) different versions of stockfish at the setup time.

PR #90, opened 2021-04-03
Add --split option to split root dirs over various sub dirs

vondele:


vondele:
@Sopel97 before merging, maybe have a look that the functionality also works for your setup. It does work for me.

PR #92, opened 2021-04-12
Make cross_eval_check script form batches from fens, delegated to the data loader.

Sopel97:
Thich means that there's no need to implement the feature transformer index generation in python code. This is particularily important for new feature blocks. Before this the features had to be generated in 3 places - data loader, player, trainer. Now the trainer never needs to know the feature indices.

I also might become useful for other things. The newly added function is `nnue_dataset.make_sparse_batch_from_fens`, which takes a feature set, list of fens, list of scores, list of ply numbers, and a list of results (in -1, 0, 1 form). Currently for each position some legal move is generated, as it doesn't really matter, but in the future the move could also be passed as an argument.

### DIFF
--- a/training_data_loader.cpp
+++ b/training_data_loader.cpp
@@ -490,6 +490,48 @@ private:
 
 extern "C" {
 
+    EXPORT SparseBatch* get_sparse_batch_from_fens(
+        const char* feature_set_c,
+        int num_fens,
+        const char* const* fens,
+        int* scores,
+        int* plies,
+        int* results
+    )
+    {
+        std::vector<TrainingDataEntry> entries;
+        entries.reserve(num_fens);
+        for (int i = 0; i < num_fens; ++i)
+        {
+            auto& e = entries.emplace_back();
+            e.pos = Position::fromFen(fens[i]);
+            movegen::forEachLegalMove(e.pos, [&](Move m){e.move = m;});
+            e.score = scores[i];
+            e.ply = plies[i];
+            e.result = results[i];
+        }
+
+        std::string_view feature_set(feature_set_c);
+        if (feature_set == "HalfKP")
+        {
+            return new SparseBatch(FeatureSet<HalfKP>{}, entries);
+        }
+        else if (feature_set == "HalfKP^")
+        {
+            return new SparseBatch(FeatureSet<HalfKPFactorized>{}, entries);
+        }
+        else if (feature_set == "HalfKA")
+        {
+            return new SparseBatch(FeatureSet<HalfKA>{}, entries);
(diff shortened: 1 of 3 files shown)


training_data_loader.cpp
```diff
@@ -490,6 +490,57 @@ struct FeaturedBatchStream : Stream<StorageT>
...
+    EXPORT SparseBatch* get_sparse_batch_from_fens(
+        const char* feature_set_c,
+        int num_fens,
+        const char* const* fens,
+        int* scores,
+        int* plies,
+        int* results
+    )
+    {
+        std::vector<TrainingDataEntry> entries;
```
vshcherbyna, line 502: Perhaps it is also a good idea to validate for repeated fens, and use std::set or something similar to avoid storing duplicates?

training_data_loader.cpp
```diff
@@ -490,6 +490,57 @@ struct FeaturedBatchStream : Stream<StorageT>
...
+        entries.reserve(num_fens);
+        for (int i = 0; i < num_fens; ++i)
+        {
+            TrainingDataEntry e;
+            e.pos = Position::fromFen(fens[i]);
+            movegen::forEachLegalMove(e.pos, [&](Move m){e.move = m;});
+            e.score = scores[i];
+            e.ply = plies[i];
+            e.result = results[i];
+            entries.push_back(std::move(e));
```
vshcherbyna: emplace_back ?

training_data_loader.cpp
vshcherbyna, line 502: Perhaps it is better to try catch the reserve method as it can throw exception and return nullptr with fprintf(stderr as it is done at the end of the function for consistency?

Sopel97: Doesn't matter, and if anything it should be done on the caller side. This is a very simple function.

training_data_loader.cpp
Sopel97: Could use emplace_back() and then use entries.back(), but this is not a performance critical part.

training_data_loader.cpp
Sopel97: If it throws we crash. If you find a realistic case where you can actually reach this (it allocates about 300 bytes per position, so you would need batches of size of millions to run out of memory) I'd be surprised. The batch itself will use a few times more memory than this.

training_data_loader.cpp
Sopel97: Okay, I guess it's simpler like that. Modified.

vshcherbyna: thanks!

PR #102, opened 2021-05-03
Bring back the old behavior for the GC being applied on the feature transformer weights.

### LINKED ISSUE #101: Ranger GC when the weights are transposed.
With the introduction of the new kernel the weights of the feature transformer are now transposed, which means ranger's GC is applied on a different axis. Maybe that's the source of the regression? Will try to disable GC and will have more data tomorrow.

This investigation follows a weird regression from vondele and weird evaluations favoring black even in startpos in my recent nets.

Sopel97:
Fixes #101

### DIFF
--- a/model.py
+++ b/model.py
@@ -136,7 +136,8 @@ class NNUE(pl.LightningModule):
     # Train with a lower LR on the output layer
     LR = 1e-3
     train_params = [
-      {'params': self.get_layers(lambda x: self.output != x), 'lr': LR},
+      {'params': self.get_layers(lambda x: self.output != x and self.input != x), 'lr': LR},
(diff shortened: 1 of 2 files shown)


Sopel97:
master vs (half of this) patch. On a different branch so loss is not comparable directly.
plot_52_53_split

vondele:
IMO, good to get this in soon. 

Probably wise to test the full version of this patch as well. If you haven't done so yet.

Sopel97:
I have not, but I will have a comparison in about 24 hours.

Sopel97:
I was incorrect. The other codepath that I have not modified for the tests is not taken with our parameters. Therefore the tests above reflect a fully fixed version.

PR #105, opened 2021-05-15
Nnue docs update

Sopel97:


glinscott:
This is really a work of art when it comes to documentation/explanations about NNUE :).

PR #106, opened 2021-05-15
.bat script for g++

aga-mp:
Cmake .bat file renamed. Added a second .bat file to compile with g++ (msys/mingw).


vondele:
note that your .bat only works on windows (because .dll instead of .so). However, I just wonder why the other script doesn't work for you. As soon as you have cmake, it should work, no?

aga-mp:
no idea, i have cmake and it did not want to work... i did not really investigate why, just went directly to msys/g++

PR #107, opened 2021-05-15
.bat script for msys/mingw

aga-mp:
Cmake .bat renamed. Added a second .bat script for g++ compile (windows msys/mingw).


aga-mp:
okay, this one should be, hopefully, without the merge points

PR #110, opened 2021-05-18
Update trainer to the new architecture.

Sopel97:
This is a slightly cleaned up version of https://github.com/Sopel97/nnue-pytorch/tree/experiment_56, with a few changes:

- weight clipping is moved outside of the optimizer, should behave about the same
- updated visualize.py to at least work, though it doesn't visualize fc layers and the plot for ft weights is slightly off
- updated cross_eval_check.py
- updated perf_sigmoid_fitter.py
- changed LR scheduling based on further results

Now will follow results from individual changes. The separate graphs are not comparable between each other.

- weight clipping 
plot_14_15
- per psqt features 
plot_20_21
- 8 subnets and psqt buckets 
plot_21_30
- MSE over WDL-space evals (not very visible on the graph, but the trend was 3 elo better. Loss not comparable)
plot_48_49
- disable GC (lower loss)
plot_53_54
- drop LR by 0.987 every epoch (trend ~2 elo better, easier parametrization) 
plot_56_57

The new architecture diagram:
HalfKAv2-45056-512x2P8x2 -16-32-1 x8

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,11 +54,11 @@ python train.py --resume_from_checkpoint <path> ...
 python train.py --gpus 1 ...
 ```
 ## Feature set selection
-By default the trainer uses a factorized HalfKP feature set (named "HalfKP^")
+By default the trainer uses a factorized HalfKAv2 feature set (named "HalfKAv2^")
 If you wish to change the feature set used then you can use the `--features=NAME` option. For the list of available features see `--help`
 The default is:
 ```
-python train.py ... --features="HalfKP^"
+python train.py ... --features="HalfKAv2^"
 ```
 
 ## Skipping certain fens in the training
@@ -69,7 +69,7 @@ python train.py ... --features="HalfKP^"
 ## Current recommended training invocation
 
 ```
-python train.py --smart-fen-skipping --random-fen-skipping 10 --batch-size 16384 --threads 8 --num-workers 8 --gpus 1 trainingdata validationdata
+python train.py --smart-fen-skipping --random-fen-skipping 3 --batch-size 16384 --threads 8 --num-workers 8 --gpus 1 trainingdata validationdata
 ```
 best nets have been trained with 16B d9-scored nets, training runs >200 epochs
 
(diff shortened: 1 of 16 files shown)


vondele:
would it make sense to make _default_feature_set_name HalfKAv2^ and the default batch size 16k as well --random-fen-skipping N and --smart-fen-skipping ? 
after that update the readme section `Current recommended training invocation`

Sopel97:
I've updated the defaults (I was using random-fen-skipping=3 and it's sufficient). As for the --smart-fen-skipping, I added --no-smart-fen-skipping that disables it, and made it enabled by default. The old option is left for compatibility but doesn't do anything.

vondele:
as we're at it, set the default number of threads and workers to 2 (or some reasonable number)? 
Similarly, what about allowing just one data file name as well (so by default training and validation is the same data set)? 

One thing I've been thinking running a little more positions for validation would give smoother curve, that's maybe without real benefit, I don't know, would ease comparison of trainings, at little extra cost.

Sopel97:
so... about threads and workers. I tried on a xeon platinum a few days ago, and I don't measure any difference between 1,2,4,8. CPU utilization is not even going higher. I think 1/1 is really the best default.

> Similarly, what about allowing just one data file name as well (so by default training and validation is the same data set)?

let's leave this for a different pr

> One thing I've been thinking running a little more positions for validation would give smoother curve, that's maybe without real benefit

The dips are at least in similar places. I can read from the graphs loss differences that result in 2-3 elo at 20k nodes per move.

vondele:
OK, let me give it a quick testing in the current state.

vondele:
```
  File "train.py", line 108, in <module>
    main()
  File "train.py", line 103, in main
    train, val = data_loader_cc(args.train, args.val, feature_set, args.num_workers, batch_size, args.smart_fen_skipping, args.random_fen_skipping, main_device)
AttributeError: 'Namespace' object has no attribute 'smart_fen_skipping'
```
for
```
python train.py ../gensfen/gensfen_random_multi_pv_diff/large_gensfen_multipvdiff_100_d9.binpack  ../gensfen/gensfen_random_multi_pv_diff/large_gensfen_multipvdiff_100_d9.binpack --gpus 1 --threads 2 --num-workers 2 --batch-size 16384 --progress_bar_refresh_rate 300 --smart-fen-skipping --random-fen-skipping 3 --lambda=1.0  --max_epochs=400
```

Sopel97:
right, sorry, fixed

PR #115, opened 2021-05-23
Simplify piece counting (non-functional change)

niklasf:


Sopel97:
I can't find the documentation for these functions, are they a stable interface?

niklasf:
Yes, it's an oversight that they are missing from the documentation. Will fix.

Btw. maybe this PR is pointless after #116. I was just playing with the code because I was curious in which ways python-chess is used here and why it is pinned to python-chess==0.31.4.

Sopel97:
> 
> 
> Yes, it's an oversight that they are missing from the documentation. Will fix.
> 
> Btw. maybe this PR is pointless after #116. I was just playing with the code because I was curious in which ways python-chess is used here and why it is pinned to python-chess==0.31.4.

Yes, it might be redundant in the future if we decide to forgo the python data loader. Your PR is reminded me that we actually still support it. I will merge anyway because it's fine.

Sopel97:
> why it is pinned to python-chess==0.31.4.

I was unable to install a later version with python 3.6.8

PR #118, opened 2021-05-24
Use C++ data loader for cross_check_eval.py

Sopel97:
This PR changes the data loader used in cross_check_eval.py from the python .bin loader to the C++ data loader (which also means it now supports .binpack). There is now a way to create a data stream that produces raw fens.

NOTE: After recent changes in official-stockfish/Stockfish/master this script needs special care to use. The nnue evaluation in the current master is never pure, and therefore there is mismatch between what the trainer does and what the player does. To get meaningful results one must simplify the additions to evaluate_nnue.cpp:evaluate


vondele:
@Sopel97 I was expecting this to be taken care off by the 'adjusted' argument of evaluate already. Do you see something else?

Sopel97:
lazy doesn't respect that it seems

PR #121, opened 2021-05-27
Remove old stuff and the python .bin data loader.

### LINKED ISSUE #116: Remove the python data loader?
Maybe we should consider removing the python data loader? It's not usable in practical terms, and drags behind the need of updating it along the c++ data loader.

Sopel97:
Closes #116


model.py
```diff
@@ -7,7 +7,7 @@
 from feature_transformer import DoubleFeatureTransformerSlice
 
 # 3 layer fully connected network
-L1 = 512
+L1 = 64
```
vondele: probably not intended ?

Sopel97: obviously lol thanks for catching that

vondele:
I think this can be merged, was waiting for @glinscott for feedback.

glinscott:
Ah, sorry for the delay, yup, lgtm.

PR #122, opened 2021-05-28
Perform a check of required files

vondele:


run_games.py
```diff
@@ -241,14 +238,34 @@ def main():
...
+    if stockfish_test is None:
+        stockfish_test = stockfish_base
+
+    if not os.path.exists(stockfish_base):
+       sys.exit("Stockfish base does not exist!")
+
+    if not os.path.exists(stockfish_test):
+       sys.exit("Stockfish test does not exist!")
+
+    if not os.path.exists(args.ordo_exe):
```
Sopel97: does this respect the `PATH` variable? In my setups I just put "ordo" and assume it can correctly found by the shell.

Sopel97: perhaps do some `ordo --version` or an equivalent?

run_games.py
```diff
@@ -241,14 +238,34 @@ def main():
...
+       sys.exit("ordo does not exist!")
+
+    if not os.path.exists(args.c_chess_exe):
```
Sopel97: same here

run_games.py
vondele: OK, good idea, need to check.

vondele: This should be resolved using `shutil.which()`

PR #129, opened 2021-06-14
Fix outcome loss

sergiovieri:
Outcome loss should be against q, the net output.


Sopel97:
embarrasing is an understatement, thanks

vondele:
OMG

PR #134, opened 2021-06-28
Move native stuff to cpp directory. Build artifacts to ./bin. Add native compression lib.

Sopel97:
This patch mainly adds [arithmetic coding] compression ability for numpy arrays. The motivation for this patch is to later introduce a new (simple data-block-based) nnue network storage format with compression support in the specification. 

An array with elements of any integral type can be compressed. The array being compressed is interpreted as either an array of `uint8_t`s or `uint16_t`s, being the symbols. Endianness is respected as explained in ./cpp/compression.cpp:109. The encoding/decoding process and the serialized format is well documented in inline comments inside `./cpp/compression.cpp`.

The python `compression` module is added (with bindings to the native C++ implementation) which exposes the following functions:
```python

'''
Destroy the array pointed to by entries ctypes object.
entries is expected to be a result of a decompress_numpy_array call.
'''
def destroy_entries(entries):
    ...
'''
Destroys the CompressedBlock pointed to by compressed_block ctypes object.
compressed_block is expected to be a result of a compress_numpy_array call
'''
def destroy_compressed_block(compressed_block):
    ...

'''
If possible, compresses the numpy array. The numpy array might
contain entries of any integer type. The array must be c_contiguous.
symbol_type must be either ctypes.c_uint8 or ctypes.c_uint16.
The function throws Exception on failure, and may fail if the compressed
size exceeds the uncompressed size.
'''
def compress_numpy_array(array, symbol_type):
    ...

'''
Decompresses a CompressedBlock pointed by compressed_block ctypes object.
entry_type must be a numpy integer dtype.
symbol_type must be either ctypes.c_uint8 or ctypes.c_uint16.
On success it return a numpy array with elements of dtype=entry_type 
and the number of elements equal to compressed_block.contents.num_entries.
Throws Exception on failure.
'''
def decompress_numpy_array(compressed_block, entry_type, symbol_type):
    ...
```

and the following class:
```python

'''
A class representing a block of compressed data that can be decompressed using
the function decompress_numpy_array. This structure does NOT have stable ABI, and
dedicated serialization format should be used. If the user wishes to decompress
a block that has not been a result of compress_numpy_array call (for example
read from a file) they can manually create a proper CompressedBlock instance.

data is the pointer to the raw compressed bytes
size is the size in bytes of the compressed data
num_entries is the number of entries that were compressed and should be decompressed.
    That is, the number of entries must be known both during compression and decompression.
'''
class CompressedBlock(ctypes.Structure):
    _fields_ = [
        ('data', ctypes.POINTER(ctypes.c_uint8)),
        ('size', ctypes.c_size_t),
        ('num_entries', ctypes.c_size_t)
    ]
```

A test can be ran by `python compression.py`

Additional changes involve restructuring of way we handle the native code and lib. All C++ code has been moved to the `./cpp` subdirectory, and the build is contained within it too. Name of the library changed to `native_bits` and a single python file loading it (`native_bits.py`) has been introduced. The location of the shared library is now the `./bin/` directory. cmake should be called within `./cpp` directory, and compilation helper script has been changed accordingly. Readme has been updated.


Sopel97:
If this is considered too much for now I'd like to at least get the part unrelated to compression merged.

vondele:
I've had little time to look at code recently, but in general adding compression might be worthwhile if the decompression can be code easily in SF.

vondele:
actually, might need to think as well on the exporting of nets from SF... that might become tricky.

Sopel97:
The same code would be included in stockfish. It contains both compression and decompression code.

PR #137, opened 2021-07-01
Add a way to disable factorization

Sopel97:
This patch adds a way to disable factorization of the L1 and the FT layer. Currently it's only possible from the code, with no option exposed to the user. A wrapper is introduced for the optimizer to handle partial layer masks; in the future weight clipping should be moved there.

### DIFF
--- a/optimizer.py
+++ b/optimizer.py
@@ -1,0 +1,81 @@
+from torch.optim.optimizer import Optimizer, required
+import torch
+
+def NnueOptimizer(optimizer_cls, params, **kwargs):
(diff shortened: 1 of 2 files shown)


optimizer.py
```diff
@@ -0,0 +1,81 @@
+from torch.optim.optimizer import Optimizer, required
+import torch
+
+def NnueOptimizer(optimizer_cls, params, **kwargs):
+    class SpecificNnueOptimizer(optimizer_cls):
```
Sopel97, line 5: if anyone has an idea how to do this part better I'd be happy to know

PR #140, opened 2021-07-08
[RFC] Initial network storage format spec.

Sopel97:
Previous attempt here: https://github.com/glinscott/nnue-pytorch/pull/64

Since then I realized that we need something simpler. So this is about as simple as we can get while preserving the required functionality. See `docs/network_storage_format.md`.


vondele:
hard to judge for me, I think the SPDX license identifier would be interesting to have.

Sopel97:
right, forgot about that one!

PR #143, opened 2021-08-09
Update trainer for 2021-08-09 new master net

Sopel97:
See https://github.com/official-stockfish/Stockfish/pull/3646

The changes are:

- HalfKAv2_hm
- 1024x2-8-32-1
- lambda applied before the loss function (interpolate score and result)
- gamma=0.992
- LR=8.75e-4
- don't reduce LR for the last layer


vondele:
looks good at first glance, will merge now

PR #148, opened 2021-08-30
Update visualization to HalfKAv2_hm-1024x2-8-32-1

Sopel97:
Since "_hm" normally has only half of the weights (for half of the king positions) it would look bad in the visualization and would be harder to make sense of, so we plot each weight twice, just like it would be used during play, to fill all 64 king squares.


hippke:
Works great for me

PR #152, opened 2021-09-16
Add GPLv3 license notice.

### LINKED ISSUE #151: Add license to repository
- I would like to fork the repository and modify some code for my own experiments. Ideally, I would like to put the result of these experiments in a different Github repository.
- Please add a license to make this valid. I suggest to use the GPL 3.0 license as for Stockfish.
- [Reason](https://opensource.stackexchange.com/questions/1720/what-can-i-assume-if-a-publicly-published-project-has-no-license): "If a repository has no license, then all rights are reserved and it is not Open Source or Free. You cannot modify or redistribute this code without explicit permission from the copyright holder."

Sopel97:
Closes #151 

It was brought to attention that this reposity does not contain any license. It would make sense to license it with GPLv3, to match Stockfish, but if there are strong objections we can consider a different license. This better be done early as it requires an agreement from all authors.

- [x] @Sopel97 
- [x] @glinscott 
- [x] @vondele 
- [x] @ddobbelaere 
- [x] @sergiovieri 
- [x] @niklasf 
- [x] @kennyfrc 
- [x] @nodchip 
- [ ] @Nomoras (only docs)

I'd appreciate every pinged person to give a clear answer below.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
+
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
+the GNU General Public License is intended to guarantee your freedom to
+share and change all versions of a program--to make sure it remains free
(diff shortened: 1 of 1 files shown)


nodchip:
I agree to license with GPLv3.

vondele:
OK with me

kennyfrc:
Ok with me

niklasf:
Ok with me

ddobbelaere:
OK with me

glinscott:
Oops, should have done this from the start.  yes, looks good to me

sergiovieri:
OK with me

vondele:
@Nomoras had a contribution to the documentation. It would be nice to have his OK, but GPL and docs is a corner case.

All code contributions are covered, we can add the license.

PR #155, opened 2021-11-29
Introduce skipping based on the likelihood of the game outcome

vondele:
this intends to skip (or more correctly weight) data points that are possibly incorrectly evaluated, i.e. retain data that are more likely to be correct.

Was used to train two recent SF nets:
https://github.com/official-stockfish/Stockfish/pull/3816
https://github.com/official-stockfish/Stockfish/pull/3808

--no-wld-fen-skipping option can be used to disable the default

### DIFF
--- a/lib/nnue_training_data_formats.h
+++ b/lib/nnue_training_data_formats.h
@@ -26,11 +26,13 @@ THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 #pragma once
 
+#include <algorithm>
 #include <cstdio>
 #include <cassert>
 #include <string>
 #include <string_view>
 #include <vector>
+#include <cmath>
 #include <memory>
 #include <fstream>
 #include <cstring>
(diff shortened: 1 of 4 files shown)


Sopel97:
Looks good. One thing that needs to be done though is that we must update the signature for [`create_fen_batch_stream`](https://github.com/glinscott/nnue-pytorch/blob/87d2a9db8b51a4269597c3ec74b88104cebbbad9/nnue_dataset.py#L73) and [`create_sparse_batch_stream`](https://github.com/glinscott/nnue-pytorch/blob/87d2a9db8b51a4269597c3ec74b88104cebbbad9/nnue_dataset.py#L172). It causes issues randomly on some linux distributions, and may have other weird effects.

I'm currently testing this approach with the original depth 9 dataset generated by stockfish, as I think this was only tested with lc0 data so far. I think it's wise to have more data, and possibly document better when this option is beneficial and when it isn't. There's no hurry.

vondele:
thanks, fixed.

The code needed to add a user changeable parameter to that part of the code is quite involved.

I'm running a training with this PR to compare to branch I effectively used, so far this looks OK. I agree testing on SF data would be quite interesting.

vondele:
I've added an additional comment, and can confirm training goes well with this branch.

Sopel97:
I have some data now. The best description is "The unreasonable effectiveness of stochastic weighting".

First, for the depth 9 data we use to train nets from scratch before retraining the resulting elo is the same as without this patch. Loss is lower, contrary to what's seen with the lc0 converted datasets.

Second, my approach to change this patch into weighting applied to loss values failed, which is surprising as they should be equivalent on average.

Below are the train loss, validation loss, and elo plots for 4 experiments.

- experiment 149 - baseline for this test, the original net I trained with the d9 data
- experiment 255 - this patch unaltered
- experiment 256 - this patch that I tried to convert into a non-stochastic version https://github.com/Sopel97/nnue-pytorch/commit/a61b1003f4e1f6f67d05c95ac89054ad7c9c7f3a
- experiment 257 - the above but without the `weight *= ...` line (no normalization)

plot_149_255_256_257

I think this patch is good to go.

vondele:
thanks for the analysis. One difference in your setup is also the testing against a fairly weak reference engine, as the Elo is about 140. This could be influencing results (the NCM result dropped by about 10 Elo after the first net with this approach, IIRC, even though it was a solid gain on fishtest) a bit. 

The result of the non-stochastic approach is somewhat surprising... need to think about that. 

I agree the patch can be merged after that (also observing good fishtest results with this branch).

PR #158, opened 2021-12-13
Use smaller initial LR

vondele:
lead to a new master net https://github.com/official-stockfish/Stockfish/pull/3848
this is in combination with a restart from a good net (previous master). Visually,
the net starts a bit better after the restart, probably logical as the initial perturbations
are smaller.

### DIFF
--- a/model.py
+++ b/model.py
@@ -295,7 +295,7 @@ class NNUE(pl.LightningModule):
 
   def configure_optimizers(self):
     # Train with a lower LR on the output layer
-    LR = 8.75e-4
+    LR = 4.375e-4
     train_params = [
       {'params' : get_parameters([self.input]), 'lr' : LR, 'gc_dim' : 0 },
       {'params' : [self.layer_stacks.l1_fact.weight], 'lr' : LR },


Sopel97:
I think we should instead add an "lr" option to the train script. The right LR may be different for the retraining process compared to training from scratch.

vondele:
yes, it would be good to have a few things more as an option. This makes it somewhat easier to test as well. Nevertheless, good defaults are needed, and I propose using those that lead to a net in master is the safe bet. 

I'll see if I can update the PR later today/this week.

Sopel97:
Perhaps we could only apply this LR change if training from model? I.e. when either `resume_from_checkpoint` or `resume-from-model` is used.

vondele:
I think too much 'smart logic' leads to unexpected behaviour, better have it explicit.

Sopel97:
I will insist at least on a comment specifying that `LR = 8.75e-4` may yield better results when training networks from scratch.

vondele:
yes, I think we should probably document the best options for training from scratch vs restarts... as said, establish the full procedure for training from scratch (which, as I understand, includes retraining cycles).

vondele:
led to another master net https://github.com/official-stockfish/Stockfish/pull/3870

However, some experiments training from scratch shows indeed that this is not optimal for that purpose, only for retraining. As such, shouldn't be merged in this form. Keeping open for the time being.

Sopel97:
superseeded by new arch trainer PR #164

PR #164, opened 2022-02-10
SFNNv4 trainer

Sopel97:
This *should* be everything, but it would be best if the results were first replicated before merging. Also, not sure how to squash this, as there's a merge commit in between.

### DIFF
--- a/model.py
+++ b/model.py
@@ -121,7 +121,7 @@ class NNUE(pl.LightningModule):
 
   It is not ideal for training a Pytorch quantized model directly.
   """
-  def __init__(self, feature_set, lambda_=1.0):
+  def __init__(self, feature_set, lambda_=1.0, gamma=0.992, lr=8.75e-4):
     super(NNUE, self).__init__()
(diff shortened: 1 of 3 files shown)


vondele:
I can try to do a few retrains once the new arch lands in SF master. It might take a couple of days before I can start, depending on timing.

scripts/run_games.sh
```diff
@@ -0,0 +1,13 @@
...
+cd ..
+
+python3 run_games.py \
+    --concurrency=24 \
+    --ordo_exe ordo \
+    --c_chess_exe c-chess-cli \
+    --features=HalfKAv2_hm^ \
+    --stockfish_base ../nnue-pytorch-training/stockfish/stockfish \
+    --stockfish_test ../nnue-pytorch-training/experiment_$1/stockfish \
+    --book_file_name ../nnue-pytorch-training/book/noob_3moves.epd \
```
vondele: you should consider switching to UHO, also for the ranking... it could make a difference.

Sopel97: I did

PR #170, opened 2022-04-03
Add -DNDEBUG to normal compile flags

vondele:
small speedup, if limited by data reading.
second commit, larger speedup (30%), using -march=native


Chaneybenjamini:
-DNDEBUG should already be set by default by line 523 in the Makefile.

I like the idea of adding -march=native, but it seems like it should be optional. Right now it is very easy to compile Stockfish for other architectures, and this makes it harder.

PR #172, opened 2022-04-17
Fix the document.

CGLemon:
Some pictures were not shown in the document. Remove some ```<span></span>``` to fix it.


Sopel97:
Thanks. I did not notice that github for some reason doesn't support <span> tags in markdown... Yet another standard feature github doesn't support.

PR #173, opened 2022-04-19
Skip positions to target a distribution

vondele:
skips positions to target a prescribed distribution of positions with number of pieces.

Uses i*(32-i)/(16*16)+1

Introduces a max skipping factor of 15.

Needs some further work to make this optional and more flexible.


Sopel97:
I think we should merge this as is, because https://github.com/official-stockfish/Stockfish/pull/4020 is based on patches on top of this.

vondele:
OK, it is correct and functional, so can be merged as is.

PR #174, opened 2022-05-12
[WIP] Add a script that allows easy combined setup, training, and testing with a single command.

Sopel97:
Currently training the nets is quite involved if one wants to have the whole setup. There's many dependencies; multiple runs require managing multiple trainer instances; network testing requires setting up a lot of binaries. To glue everything together one must know the directory trees generated and consumed by the programs, which is often not trivial. Resulting artifacts need to be managed.

The purpose of this PR is to add a single standalone script that handles the environment setup, network training, network testing, and artifact collection. Even though it's distributed with the whole repo it can be used alone, and that is the most important use-case, since it sets up nnue-pytorch itself anyway. It should make it possible to run the full network training setup either automatically or by people who are not developers.

Currently the following features are implemented:

- verification of most of the dependencies (still requires user intervention)
- directory tree setup with distinction of self-contained experiments
- extensive parametrization for cli invocation
- automatic ordo setup
- automatic c-chess-cli setup
- automatic nnue-pytorch setup from given repo link (branch or commit) + data loader compilation
- automatic setup of baseline and tested engine from given repo links (branch or commit)
- automatic opening book download and setup for URL-based books
- automatic (optional) network training process with distinctions into runs and support for using multiple gpus (at most 1 gpu per run, trainer limitation), single session
- automatic (optional) network testing process (run_games.py functionality)
- optional nice TUI dashboard
- or simple streamed output so that it's possible to run as a subprocess
- safety regarding detached stray processes

Dashboard example (slightly out of date): 
trainer

The basic requirements to start are:

- python >=3.7
- gcc, g++ (mingw on windows)

What still needs to be done.

- **testing on linux** (currently only tested on windows 7)
- revision of error handling (and what to actually do on errors)
- some code cleanup and restructure
- comments
- revision of the output for the mode without TUI, it's very crude right now
- handling trainer errors (like cuda OOM for example)
- logging
- packaging resulting artifacts (input=cli invocation, output=nets+logs)
- possibly verify all nets with cross_check_eval.py as they are produced
- possibly do some real time real time plots
- possibly a web-based UI

Additional (mostly necessary) changes made to other modules:

- train.py
  - allow varying epoch/validation size (useful for testing purposes)
  - parametrized network checkpointing
  - workaround for https://github.com/pytorch/pytorch/issues/48888
- run_games.py
  - parametrized time control, hash, threads
  - parametrized game batch size
  - change serialize.py invocation to subprocess.Popen to prevent "^" in feature set name from breaking stuff on windows
  - update resign and draw c-chess-cli parameters to the new version (a better solution would be to attempt both ways but needpessly verbose. let's just move on.
  - omit stdbuf for c-chess-cli on unsupported systems (win32)
  - flush stdout after each line from processed c-chess-cli output to make it easier to digest from a parent process
  - run ordo in parallel to matches (ordo is mostly single threaded anyway and for longer runs it can take more than a few minutes to run in my experience)


vondele:
Some things needed to get it to work on linux, just a dump of my notes for now:
```
adjust easy_train_example.bat to easy_train_example.sh
* deal with ^ vs \ and escaping of ^
* add data sets where they are
* add threads and workers correctly

pip install psutils            (not detected)
pip install asciimatics        (detected)
pip install GPUtil             (detected)

fail on UHO book download (URL escape)

direct execution of bash script (should be passed to bash)
* with subprocess.Popen([os.path.join(directory, 'compile_data_loader.bat')], cwd=directory) as process:

Output in TUI a bit screwed.

```

Sopel97:
Thanks. This will be helpful once I start looking into linux support.
It all makes sense beside one thing

> pip install psutils            (not detected)

did you mean `pip install psutil`? (also I thought that it's in stdlib, but apparently not, should be checked for on startup then)

vondele:
`pip install psutil` indeed.

vondele:
Gui view 
image

Sopel97:
hmm, so it gets overwritten by the normal console output

vondele:
I guess the 'no games' state is because the example uses 
```
    --engine-base=official-stockfish/Stockfish/master \
    --engine-test=official-stockfish/Stockfish/master \
```
so not a functional player for this net arch. What's the link for a working player?

vondele:
networks are produced. I get this error message:
```
  File "run_games.py", line 111, in run_match
    process = subprocess.Popen("stdbuf -o0 " + command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
  File "/usr/lib/python3.8/subprocess.py", line 858, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.8/subprocess.py", line 1704, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'stdbuf -o0 /home/vondele/chess/vondele/nnue-pytorch/scripts/easy_train_data/c-chess-cli/c-chess-cli -each tc=10000+10000 nodes=1000 option.Hash=8 option.Threads=1 -gauntlet -games 200 -rounds 1 -concurrency 2 -openings file=/home/vondele/chess/vondele/nnue-pytorch/scripts/easy_train_data/experiments/experiment_test/books/UHO_XXL_+0.90_+1.19.epd order=random -repeat -resign count=3 score=700 -draw count=8 score=10 -engine cmd=/home/vondele/chess/vondele/nnue-pytorch/scripts/easy_train_data/experiments/experiment_test/stockfish_base/src/stockfish name=master -engine cmd=/home/vondele/chess/vondele/nnue-pytorch/scripts/easy_train_data/experiments/experiment_test/stockfish_test/src/stockfish name=/home/vondele/chess/vondele/nnue-pytorch/scripts/easy_train_data/experiments/experiment_test/training/run_0/nn-epoch1.nnue option.EvalFile=/home/vondele/chess/vondele/nnue-pytorch/scripts/easy_train_data/experiments/experiment_test/training/run_0/nn-epoch1.nnue -engine cmd=/home/vondele/chess/vondele/nnue-pytorch/scripts/easy_train_data/experiments/experiment_test/stockfish_test/src/stockfish name=/home/vondele/chess/vondele/nnue-pytorch/scripts/easy_train_data/experiments/experiment_test/training/run_0/nn-epoch5.nnue option.EvalFile=/home/vondele/chess/vondele/nnue-pytorch/scripts/easy_train_data/experiments/experiment_test/training/run_0/nn-epoch5.nnue -engine cmd=/home/vondele/chess/vondele/nnue-pytorch/scripts/easy_train_data/experiments/experiment_test/stockfish_test/src/stockfish name=/home/vondele/chess/vondele/nnue-pytorch/scripts/easy_train_data/experiments/experiment_test/training/run_0/nn-epoch8.nnue option.EvalFile=/home/vondele/chess/vondele/nnue-pytorch/scripts/easy_train_data/experiments/experiment_test/training/run_0/nn-epoch8.nnue -pgn /home/vondele/chess/vondele/nnue-pytorch/scripts/easy_train_data/experiments/experiment_test/training/out_temp.pgn 0'
```

Sopel97:
Popen might be thinking that it's one large exectutable path...

vondele:
yes, it looks like that.

ppigazzini:
> args is required for all calls and should be a string, or a sequence of program arguments. Providing a sequence of arguments is generally preferred, as it allows the module to take care of any required escaping and quoting of arguments (e.g. to permit spaces in file names). If passing a single string, either shell must be [True](https://docs.python.org/3.10/library/constants.html#True) (see below) or else the string must simply name the program to be executed without specifying any argum

https://docs.python.org/3.10/library/subprocess.html#frequently-used-arguments

ppigazzini:
Not a priority but the code can be simplified using `pathlib` (multi OS too), some examples:
- from:
```python
def convert_ckpt(root_dir,features):
    """ Find the list of checkpoints that are available, and convert those that have no matching .nnue """
    # run96/run0/default/version_0/checkpoints/epoch=3.ckpt, or epoch=3-step=321151.ckpt
    p = re.compile("epoch.*\.ckpt")
    ckpts = []
    for path, subdirs, files in os.walk(root_dir, followlinks=False):
        for filename in files:
            m = p.match(filename)
            if m:
                ckpts.append(os.path.join(path, filename))
```
- to:
```python
def convert_ckpt(root_dir,features):
    """ Find the list of checkpoints that are available, and convert those that have no matching .nnue """
    # run96/run0/default/version_0/checkpoints/epoch=3.ckpt, or epoch=3-step=321151.ckpt
    # pathlib glob does not follow symlinks
    ckpts = [file for file in Path(root_dir).rglob("epoch*.ckpt")]
```
- from:
```python
    main_pgn_file_name = os.path.join(root_dir, "out.pgn")
    curr_pgn_file_name = os.path.join(root_dir, "out_temp.pgn")
    if not os.path.exists(main_pgn_file_name):
        with open(main_pgn_file_name, 'w'): pass
    try:
        with open(main_pgn_file_name, 'a') as file_to:
            with open(curr_pgn_file_name, 'r') as file_from:
                for line in file_from:
                    file_to.write(line)
    except:
        print('Something went wrong when adding new games to the main file.')
```
- to:
```python
    root_dir = Path("something")
    main_pgn_file_name = root_dir / "out.pgn"
    curr_pgn_file_name = root_dir / "out_temp.pgn"

    try:
        with main_pgn_file_name.open("a") as file_to:
            file_to.write(curr_pgn_file_name.read_text())
    except Exception:
        print('Something went wrong when adding new games to the main file.')
```

Sopel97:
Changes:

- run ordo with subprocess instead of os.system
- run c-chess-cli correctly (hopefully) on linux
- run data loader compilation through sh on linux
- catch trainer stderr so that hopefully there's no more overwrite on the TUI in some terminals
- handle CUDA OOM
- use rglob instead of os.walk as suggested

To take effect one must delete the previous experiment directory. Normally it is required to run the script with empty directory to let it setup everything in entirety, but the example script overrides this with `--fail-on-experiment-exists=False` which is useful for debugging but prevents some setup from reoccuring.

Sopel97:
Currently the behaviour of --fail-on-exerpiment-exists=False is that it starts from scratch, just creates a new subdirectory (version_[\d+]) for the networks (default PL behaviour). I was considering an option to restart from the latest checkpoint (utilizing --resume-from-checkpoint) in each run, but I'm not sure how safe it is for reproducibility, since we have no way to resume reading from the exact same place in the dataset. If this is considered not an issue, or enough to document possible issues, then I'll add this functionality.

vondele:
I think that having such an option would be great (I for one would need that, I can't run the trainer for more than say 1day). 

I think we can't have exact binary reproducibility of the training anyway, that's not something pytorch (on CUDA) guarantees (floating point reductions etc are not ordered, I assume).

vondele:
the current script seems to work on linux, there is still some output overwrite the TUI.
image

I'll will try to make a PR to your branch with my example script that makes it easier to test for me.

vondele:
minor suggestion, what about moving the books directoy a little higher in the tree (like next to c-chess-cli), so that it won't be do downloaded for each new experiment?

vondele:
The output overwrite is due to this line:
```
diff --git a/scripts/easy_train.py b/scripts/easy_train.py
index 344d05e..c9ce47b 100644
--- a/scripts/easy_train.py
+++ b/scripts/easy_train.py
@@ -484,8 +484,8 @@ class TrainingRun(Thread):
                         self._has_started = True
                         if self._current_epoch == self._num_epochs - 1 and self._current_step_in_epoch >= self._num_steps_in_epoch:
                             self._has_finished = True
-                        if self._current_step_in_epoch % 100 == 0:
-                            print(line)
+                        # if self._current_step_in_epoch % 100 == 0:
+                        #     print(line)
                 except:
                     pass
                 if 'CUDA_ERROR_OUT_OF_MEMORY' in line or 'CUDA out of memory' in line:

```
is that one intentional?

Sopel97:
Okay. I see what the overwrite now comes from. This is just one place, of many. Initially I thought it was coming from subprocesses but it's actually normal stdout/stderr. I'll need to supress all console output for TUI mode.

Regarding book directory, yes, that's a mistake on my part.

Sopel97:
I've added proper logging and now there should be no writing over the TUI. Also moved the book directory.

vondele:
Thanks, I've rebased my small fixes

Sopel97:
I've added the restart functionality that we talked about. `--resume-training=True` will make the script attempt to restart from the latest checkpoint found for each run. Also added your fixes, thanks.

vondele:
Thanks, for fixing things. For me currently it/s seems to be not correct, not yet sure where the issue is.

Sopel97:
I think that's all that's important for now, at least for the functionality. I'd like you to test the new approximate ordo calculator (on by default)

scripts/easy_train.py
```diff
@@ -1647,16 +1647,18 @@ def prepare_start_model_from_experiment(directory, experiment_path, run_id, nnue
         raise Exception('Could not find any viable .ckpt nor .nnue files in the start experiment.')
     return prepare_start_model(directory, best_model, run_id, nnue_pytorch_directory, features)
 
+def get_default_feature_set_from_nnue_pytorch(nnue_pytorch_directory):
+    with open(os.path.join(nnue_pytorch_directory, 'features.py'), 'r') as features_file:
+        for line in features_file:
+            line = line.strip()
+            if line.startswith('_default_feature_set_name'):
+                return line.split()[-1][1:-1]
+
```
vondele, line 2257: should we throw an exception if nothing found?

scripts/easy_train.py
```diff
@@ -0,0 +1,1844 @@
...
+    parser.add_argument("--network-testing-time-increment-per-move", type=float, default=None, dest='network_testing_time_increment_per_move', help="Number of seconds added to clock per move")
+    parser.add_argument("--network-testing-nodes-per-move", type=int, default=None, dest='network_testing_nodes_per_move', help="Number of nodes per move to use for testing. Overrides time control. Should be used ove time control for better consistency.")
+    parser.add_argument("--network-testing-hash-mb", type=int, default=8, dest='network_testing_hash_mb', help="Number of MB of memory to use for hash allocation for each engine being tested.")
+    parser.add_argument("--network-testing-games-per-round", type=int, default=20 * default_testing_threads, dest='network_testing_games_per_round', help="Number of games per round to use. Essentially a testing batch size. By default uses larger batches with larger --network-testing-threads")
+    parser.add_argument("--resume-training", type=str2bool, default=False, dest='resume_training', help="Attempts to resume each run from its latest checkpoint.")
```
vondele: I wonder if we should make this resume-training True by default, really useful from my point of view (easily quit training and resume, a real life-saver).

scripts/easy_train.py
```diff
@@ -0,0 +1,1844 @@
...
+            return
+
+    ordo_directory = os.path.join(absolute_workspace_path, 'ordo')
+    c_chess_cli_directory = os.path.join(absolute_workspace_path, 'c-chess-cli')
+    books_directory = os.path.join(absolute_workspace_path, 'books')
+
+    stockfish_base_directory = os.path.join(experiment_directory, 'stockfish_base')
+    stockfish_test_directory = os.path.join(experiment_directory, 'stockfish_test')
+    nnue_pytorch_directory = os.path.join(experiment_directory, 'nnue-pytorch')
+    bookkeeping_directory = os.path.join(experiment_directory, 'bookkeeping')
```
vondele: maybe this directory could be 'logging' as that is what we store there right now?

scripts/easy_train.py
```diff
@@ -0,0 +1,1844 @@
...
+        time_per_game=args.network_testing_time_per_game,
+        time_increment_per_move=args.network_testing_time_increment_per_move,
+        nodes_per_move=args.network_testing_nodes_per_move,
+        hash=args.network_testing_hash_mb,
+        games_per_round=args.network_testing_games_per_round,
+        active=do_network_testing,
+        additional_args=[arg for arg in args.additional_testing_args or []]
+    )
+
+    #TODO it would be nice to be able to set a time limit at the cmdline and stop after e.g. 20h
```
vondele: This would be a real nice feature for me, i.e. a clean shutdown (like 'quit') after a specified time.

scripts/easy_train.py
```diff
@@ -0,0 +1,1844 @@
...
+                    break
+                else:
+                    print('Type `quit` to stop.')
+            except EOFError:
+                # For non-interactive environments
+                time.sleep(1)
+
+    for tr in training_runs:
+        tr.stop()
+    network_testing.stop()
```
vondele, line 2448: I believe network testing continues even if training has reached max-epoch, for me it would seem natural to stop testing after training reached max-epoch.

Sopel97: stopping it at the same time results in latest networks being undertested. Usually I run testing for an hour or so after the training finishes. Not sure how to handle that best.

vondele:
The only feature I was wondering about is if we could have a one liner to train from scratch. Something like
``` python easy_train.py {options for run one} &&  python easy_train.py {options for run two}```
with the proper return value of easy_train.py this chaining should work (returning 1 if we successfully reach max-epoch, and 0 otherwise).

**Comment by vondele** [comment]
I will be able to test tonight, but at first sight this looks good.

I would like this to be my default manner of training as well.... I will usually run without the TUI, so good (extensive/complete) will be key.

File: scripts/easy_train.py (hunk 5 above)
Comment (vondele) [reply scripts/easy_train.py:2448]: Not sure either. I usually stop testing at the same time, meaning the last trained network is ignored from a testing point of view. It would be nice to have good termination criterion of easy_train. It could be max-epoch, which is rather simple, but maybe we need something more elaborate?

File: scripts/easy_train.py (hunk 4 above)
Comment (Sopel97) [reply scripts/easy_train.py]: ok. Added experimentally `--auto_exit_timeout` and `--auto_exit_timeout_on_training_finished`. Can't test right now because doesn't work on windows (needs SIGALRM).

Comment (Sopel97) [reply scripts/easy_train.py]: one thing I worry about is that it may get swallowed by some `except` if thrown in a bad moment (for example when _make_run_text is running). We'd need to explicitely catch every exception to make this work 100% of the time.

Comment (Sopel97) [reply scripts/easy_train.py]: alternatively, reraise every 1 second? (maybe longer, to allow for graceful exit?)

Comment (vondele) [reply scripts/easy_train.py]: I haven't looked at the implementation, but how does pressing 'quit' on the TUI work? Can't this mechanism be recycled for the timeout ? 

It should be graceful if possible, I don't think we care on accuracy very much (like the timeout might be 23h, and 5min doesn't matter).

**Comment by vondele** [comment]
minor nit 
image
the final step seems off by one (63/62). 

I'm trying to use `--auto-exit-timeout 100` but that doesn't seem to work. At least the TUI doesn't quite and continues. However, after quiting there is a message `Schedule exit after 100s` in the console, so I would expect it partially works.

**Comment by Sopel97** [comment]
The tui might be setting some signal handlers... It happens after our setup. The counter is wrong indeed, since it apparently doens't show step id but number of completed steps.

**Comment by vondele** [comment]
let me try without tui...

otherwise, it might be easiest to just create a running thread that checks the clock every 1s and goes to sleep in between.

**Comment by Sopel97** [comment]
Yes, we still need to use some signal though, as otherwise there's no way to get an exception in the main thread.

**Comment by vondele** [comment]
I see. However, also with --tui=False, timeout doesn't seem to work.

**Comment by vondele** [comment]
meanwhile the testing run I had finished as
```
  File "easy_train.py", line 1938, in <module>
    main()
  File "easy_train.py", line 1899, in main
    Screen.wrapper(app, catch_interrupt=True, arguments=[last_scene, training_runs, network_testing])
  File "/home/vondele/chess/vondele/nnue-pytorch/env/lib/python3.8/site-packages/asciimatics/screen.py", line 1433, in wrapper
    return func(screen, *arguments)
  File "easy_train.py", line 1477, in app
    screen.play(scenes, stop_on_resize=True, start_scene=scene, allow_int=True)
  File "/home/vondele/chess/vondele/nnue-pytorch/env/lib/python3.8/site-packages/asciimatics/screen.py", line 1606, in play
    self.draw_next_frame(repeat=repeat)
  File "/home/vondele/chess/vondele/nnue-pytorch/env/lib/python3.8/site-packages/asciimatics/screen.py", line 1709, in draw_next_frame
    effect.update(self._frame)
  File "easy_train.py", line 1449, in update
    self._update_network_testing_status()
  File "easy_train.py", line 1443, in _update_network_testing_status
    self._network_testing_status.value = self._network_testing.get_status_string()
  File "easy_train.py", line 1061, in get_status_string
    f'{self._current_test.elo:0.1f}±{self._current_test.elo_error_95:0.1f} Elo'
AttributeError: 'NoneType' object has no attribute 'elo_error_95'
```

and I have another error... probably we need to run the script a few times to completion without errors before merging.

Sopel97:
> AttributeError: 'NoneType' object has no attribute 'elo_error_95'

That's the race I didn't fix earlier. Should be fixed now.

> this probably should be made more tolerant

Fixed

Also made a scheduled exit based on os._exit, which appears to be the only viable option sadly... Though it shouldn't be much of an issue since the cleanup happens from a separate process and the easy_train script itself doesn't do anything that can leave persistent partial results.

Sopel97:
Looking at the code of asciimatics it looks like they have different codepaths for windows and linux here... On linux it tries to reset signal handlers which fails because it cannot be done on a non-main thread. Since we terminate the process soon anyway I think this is redundant, so I just copied the rest of the cleanup code. It *should* work on linux too now.

vondele:
yes, works now. I think we have a reached a good state now. Time to merge, and improve afterwards.

PR #175, opened 2022-05-16
Update trainer to match new SFNNv5 network.

Sopel97:


vondele:
This looks good to me. Do you plan to update the documentation / figure to reflect the SFNNv5 arch ?

Sopel97:
yes, there's quite a few things missing from the docs that I want to document in the following days

PR #176, opened 2022-05-17
Document Stockfish evaluation network history.

Sopel97:
First part of the documentation updates. This one involves diagrams and commit links for all Stockfish architectures.


Sopel97:
I'm also thinking about a good way to describe quantization in these networks. I wonder if a more detailed diagram (at least for these historical architectures section, where they should be described in detail) like the following would be better:

HalfKP-40960-256x2-32-32-1_withquant drawio

vondele:
I think it is good to document the quantization (maybe it is on the edge of being too detailed, ultimately not all info can be conveyed in a picture). I see a few wrong `FP[]` instead of `FP<>`.

PR #177, opened 2022-05-17
Change pngs to svgs in the docs.

Sopel97:
Does nnue.md render fine for everyone? I Notice that when opening individual svgs (from url) the browser blocks some inline images, but that shouldn't be an issue.


vondele:
image

yes, this is not rendered fully correctly (final summation is missing). It looks OK in the nnue.md doc.

as a minor nit, I would order the network architectures at the end in reverse chronological order, most recent first.

PR #178, opened 2022-05-18
Add detailed diagrams for historical Stockfish networks.

Sopel97:
These diagrams include complete quantization specification. I tried to make these as detailed as necessary while keeping everything compact. 

I also included a brief description of the feature set, that should be understandable to anyone who knows how feature sets in NNUE generally look like. I feel like it's necessary to fulfill the purpose of these diagrams. For now I included files for both diagrams - including feature set description and not including it. Displayed in nnue.md are the ones with the description

I'm open to suggestions. There might also be small typos/copy-paste-errors.


Sopel97:
Fixed a few typos and improved the feature set descriptions

PR #179, opened 2022-05-24
Fix serialize/deserialize for SFNNv5

Sopel97:
@vondele could you please test quickly if this fixes the issue?


vondele:
yes, seems to fix the exception.

PR #181, opened 2022-05-25
Various changes and improvements to easy train script.

Sopel97:


Sopel97:
- More complete logging.
- Add timeout of 20 for each engine in c-chess-cli just to be safe
- Retry running matches with c-chess-cli up to 2 times before printing "Error running matches"
- System-wide lock for initialization (for an experiment). Tested on windows, not tested on linux.

vondele:
testing on linux reveals an issue if starting without any directory structure present:
```
  File "easy_train.py", line 2568, in <module>
    main()
  File "easy_train.py", line 2374, in main
    with SystemWideMutex(os.path.join(experiment_directory, '.lock')) as mutex:
  File "easy_train.py", line 399, in __init__
    self.file = open(self.name, 'a+')
FileNotFoundError: [Errno 2] No such file or directory: '/home/vondele/chess/vondele/nnue-pytorch/scripts/easy_train_data/experiments/experiment_001/.lock'
Exception ignored in: <function SystemWideMutex.__del__ at 0x7fc9538cf940>
  File "easy_train.py", line 417, in __del__
    self.close()
  File "easy_train.py", line 410, in close
    if self.file is None:
AttributeError: 'SystemWideMutex' object has no attribute 'file'
```
after creating the directory by hand:
```
mkdir -p /home/vondele/chess/vondele/nnue-pytorch/scripts/easy_train_data/experiments/experiment_001/
```
things seem to work (as in setting up the needed tools and starting the testing). However, following a `quit` in the tui, it finishes with:
```
Exception ignored in: <function SystemWideMutex.__del__ at 0x7ffb913b2940>
  File "easy_train.py", line 417, in __del__
    self.close()
  File "easy_train.py", line 412, in close
    os.unlink(self.file)
TypeError: unlink: path should be string, bytes or os.PathLike, not _io.TextIOWrapper
```

Sopel97:
Ok. The first error makes sense, the lock needs to be somewhere else. Should be fixed now.

The second error is because os.unlink takes the path instead of a file. Should also be fixed now.

vondele:
still needs an `mkdir -p /home/vondele/chess/vondele/nnue-pytorch/scripts/easy_train_data/`

vondele:
the quit now works fine

Sopel97:
> still needs an `mkdir -p /home/vondele/chess/vondele/nnue-pytorch/scripts/easy_train_data/`

Right, I see now. Fixed.

vondele:
OK, now it works. Maybe you want to squash some of the commits before merging? If so, please go ahead.

PR #189, opened 2022-05-30
Optional local MSYS2 setup.

Sopel97:
Potentially fixes #189.

### DIFF
--- a/scripts/easy_train.py
+++ b/scripts/easy_train.py
@@ -12,6 +12,8 @@ import argparse
 import math
 import logging
 import time
+import platform
+import os
 
 EXITCODE_OK = 0
 EXITCODE_MISSING_DEPENDENCIES = 2
@@ -24,6 +26,8 @@ LOGGER.setLevel(logging.DEBUG)
 LOGGER.addHandler(logging.StreamHandler(stream=sys.stdout))
 LOGGER.propagate = False
 
+ENV = os.environ.copy()
+
 def validate_python_version():
     if sys.version_info >= (3, 7):
         LOGGER.info(f'Found python version {sys.version}. OK.')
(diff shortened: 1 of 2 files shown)


Sopel97:
A different solution to this will be made ~tomorrow. See discussion in the linked issue.

scripts/easy_train.py
```diff
@@ -1035,7 +1055,8 @@ def setup_c_chess_cli(directory):
     with open(os.path.join(directory, 'make.py', '-c', 'gcc'), 'w') as makefile:
```
ppigazzini: This seems wrong, `-c` and `gcc` should be command options, not folders.

Sopel97: pretty dumb, fixed

Sopel97:
Okay. So this is the current solution (Windows only, linux not impacted):

Added `--msys2-path` CLI argument that needs to be specified if the user wants to use an existing msys2 installation. PATH will be updated with .../msys2/, .../msys2/mingw64/bin, .../msys2/usr/bin - but I'm not actually certain that this environment will be propagated all the way through to the required subprocesses... This needs testing.

Otherwise `--allow-choco` must be set to True (just so we don't install a lot of stuff without user's consent), and the script will attempt to install Chocolatey and later MSYS2 through it. This is a local installation in the workspace directory, without automatically setting the PATH. The PATH will be setup in the local environment the same as in the case of `--msys2-path`, in which case WE MAY HAVE A PROBLEM if it doesn't work. Needs testing.

Additionally, choco requires TLS 1.2 and powershell 3.0, which are not always present on windows 7 and may require additional setup. So Windows 7 users will likely need to install msys2 manually and use `--msys2-path`.

@ppigazzini Could you perhaps test this in your windows sandbox?

Sopel97:
Do we want to maintain this or should we rely on the user to install MSYS?

PR #199, opened 2022-06-21
[WIP] Attempt to optimize the position skipping to fit a distribution.

Sopel97:
The old algorithm work online and therefore might discard excessive amounts of positions.
The new algorithm filters the buffers that are filled in each worker thread
(currently each buffer is around 1M elements). This offline algorithm allows more exact
distribution match and less wasted positions. It is however tied to the implementation of the buffering,
which is not done for .bin files (so it currently is not implemented for .bin files), and the buffer size will
now have (a small) effect. Similarly FEN stream filtering no longer uses this type of skipping.

### DIFF
--- a/lib/nnue_training_data_formats.h
+++ b/lib/nnue_training_data_formats.h
@@ -7560,12 +7560,14 @@ namespace binpack
             int concurrency,
             std::string path,
             std::ios_base::openmode om = std::ios_base::app,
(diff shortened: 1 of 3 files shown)


Sopel97:
Appears to be bugged currently in some way. Testing shows networks degrade during training which is wrong.

PR #209, opened 2022-10-07
Fix the incorrect weight scale used in NNUEReader

dhbloo:
It seems that an incorrect weight scale for output fc layers is used in NNUEReader, which causes a broken pt converted from a nnue file:

https://github.com/glinscott/nnue-pytorch/blob/77d79192877513878adc38aa4a3e9ce681be5f7f/serialize.py#L193-L195

According to NNUEWriter, this should be: 

https://github.com/glinscott/nnue-pytorch/blob/77d79192877513878adc38aa4a3e9ce681be5f7f/serialize.py#L102-L109

### DIFF
--- a/serialize.py
+++ b/serialize.py
@@ -190,8 +190,12 @@ class NNUEReader():
     layer.weight.data = torch.cat([weights, psqt_weights], dim=1)
 
   def read_fc_layer(self, layer, is_output=False):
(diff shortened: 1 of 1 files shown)


Sopel97:
It indeed appears that this part was incorrect. Probably went unnoticed because it's rarely used. Thanks for catching that.

PR #210, opened 2022-10-12
Fix the incorrect NUM_PT in HalfKAv2_hmFactorized

dhbloo:
Fixed the incorrect NUM_PT bug, as mentioned by @Pikacat-OuO in SF discord [here](https://discord.com/channels/435943710472011776/733545871911813221/1029376219415851018).


Sopel97:
For the record, this issue was not affecting current code, but may have affected code that extends upon the factorized HalfKAv2_hm features.

PR #213, opened 2022-11-25
Fix Trainer.root_gpu crash when using pytorch-lightning 1.8+

linrock:
Fixes a crash upon training start when newer versions of pytorch-lightning are used due to `Trainer.root_gpu` being removed in 1.8+ a few weeks ago.

```
AttributeError: `Trainer.root_gpu` was deprecated in v1.6 and is no longer
accessible as of v1.8. Please use `Trainer.strategy.root_device.index` instead.
```

Support for `Trainer.root_gpu` was removed in:
https://github.com/Lightning-AI/lightning/issues/11994

More info about pytorch-lightning 1.8.0 in the release notes:
https://github.com/Lightning-AI/lightning/releases/tag/1.8.0

Fixes https://github.com/glinscott/nnue-pytorch/issues/212

### DIFF
--- a/train.py
+++ b/train.py
@@ -126,7 +126,7 @@ def main():
   checkpoint_callback = pl.callbacks.ModelCheckpoint(save_last=args.save_last_network, every_n_epochs=args.network_save_period, save_top_k=-1)
   trainer = pl.Trainer.from_argparse_args(args, callbacks=[checkpoint_callback], logger=tb_logger)
 
-  main_device = trainer.root_device if trainer.root_gpu is None else 'cuda:' + str(trainer.root_gpu)
+  main_device = trainer.root_device if trainer.strategy.root_device.index is None else 'cuda:' + str(trainer.strategy.root_device.index)
 
   nnue.to(device=main_device)
 


Sopel97:
Requires checking on older versions of pytorch before merging.

vshcherbyna:
when applying this patch and when using cpu trainer (I omit "-gpus 1" parameter) I am getting the following error:

```
  File "... \nnue-pytorch\train.py", line 129, in main
    main_device = trainer.root_device if trainer.strategy.root_device.index is None else 'cuda:' + str(trainer.strategy.root_device.index)
AttributeError: 'Trainer' object has no attribute 'root_device'
```

linrock:
> when applying this patch and when using cpu trainer (I omit "-gpus 1" parameter) I am getting the following error:
> 
> ```
>   File "... \nnue-pytorch\train.py", line 129, in main
>     main_device = trainer.root_device if trainer.strategy.root_device.index is None else 'cuda:' + str(trainer.strategy.root_device.index)
> AttributeError: 'Trainer' object has no attribute 'root_device'
> ```

does it also crash this way without this patch when you use the cpu trainer without gpus? this patch didn't change the `trainer.root_device` call.

what pytorch-lightning version are you using? ie. `pip show pytorch-lightning`

ThatGuyCalledJesse:
I have this isue too, my Pytorch-lightning version is 1.7.7

Sopel97:
You need to explicitly specify a single gpu. Most straightforward by `--gpus=1`. Also see https://github.com/glinscott/nnue-pytorch/wiki/Hardware-requirements

ThatGuyCalledJesse:
Now I have this issue
``
pytorch_lightning.utilities.exceptions.MisconfigurationException: CUDAAccelerator can not run on your system since the accelerator is not available. The following accelerator(s) is available and can be passed into `accelerator` argument of `Trainer`: ['cpu'].
``
Is this an issue in the BIOS or something?

ThatGuyCalledJesse:
ah ok thanks, i'll have a look

ThatGuyCalledJesse:
It's not relevant to this repo, but do I need pip or conda to install pytorch GPU?

ThatGuyCalledJesse:
because I can't find it

Sopel97:
https://pytorch.org/

ThatGuyCalledJesse:
yeah this worked

linrock:
> Requires checking on older versions of pytorch before merging.

Confirmed training works when using the oldest installable version of torch with cuda support via pip (1.11.0+cu102)

Otherwise, the torch website's official instructions for installing very old versions of torch such as 1.7.0 don't work:
https://pytorch.org/get-started/previous-versions/#linux-and-windows-23

image

vondele:
This PR was also fixed the issue for me, I assume this can be merged?

vondele:
I'm still a bit confused. Things seem to train on GPU with this branch, but the easy_train.py script errors out:
'Found torch without CUDA but CUDA support required. Exiting'
The pkg.version string is `1.13.1` which indeed doesn't contain `cu` ... has this version string changed?

Sopel97:
We could probably replace that check with `torch.cuda.is_available()`.

vondele:
let me try that.

PR #214, opened 2022-11-25
Fix warnings about thread.setDaemon() in python 3.10+

linrock:
camelCase methods in threading were deprecated in: https://github.com/python/cpython/pull/25174

```
easy_train.py:316: DeprecationWarning: setDaemon() is deprecated,
set the daemon attribute instead thread.setDaemon(True)
```

### DIFF
--- a/scripts/easy_train.py
+++ b/scripts/easy_train.py
@@ -313,7 +313,7 @@ def schedule_exit(timeout_seconds, errcode):
         os._exit(errcode)
 
     thread = Thread(target=f)
-    thread.setDaemon(True)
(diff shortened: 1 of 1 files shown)


Sopel97:
Requires checking on older versions of python before merging. We currently require at least python 3.7 https://github.com/glinscott/nnue-pytorch/blob/5b2de095335bc0fe31921078fc6d51b5e0a42295/scripts/easy_train.py#L28

linrock:
Confirrmed working on python 3.7:

image

Setting this attribute has been available since at least 2.x

image

PR #219, opened 2022-12-21
Skip positions with score equal to Stockfish's VALUE_NONE (32002)

Sopel97:
With the recent new network for Stockfish it appears that we need to do some more costly filtering on the datasets (https://github.com/official-stockfish/Stockfish/pull/4295). Removing positions from the dataset degrades compression, so in some cases it might be better to mark positions with a special flag and skip them on load. I think it's reasonable to use `score==VALUE_NONE` as such a flag and this PR implements it.

### DIFF
--- a/training_data_loader.cpp
+++ b/training_data_loader.cpp
@@ -820,6 +820,12 @@ std::function<bool(const TrainingDataEntry&)> make_skip_predicate(bool filtered,
             wld_filtered
             ](const TrainingDataEntry& e){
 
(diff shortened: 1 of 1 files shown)


vondele:
I think this is useful.. don't know if this will help linrock with the current dataset, as he would have to recompute this from scratch, to add the positions... I assume.

Sopel97:
Yes, the dataset would have to be redone.

PR #222, opened 2022-12-30
introduce skip early plies parameter

Disservin:
convert the static value used in https://github.com/official-stockfish/Stockfish/pull/4314 into a parameter

would be cool if someone could test this since I have no idea how to run pytorch ^^

### DIFF
--- a/train.py
+++ b/train.py
@@ -10,13 +10,13 @@ from torch import set_num_threads as t_set_num_threads
 from pytorch_lightning import loggers as pl_loggers
 from torch.utils.data import DataLoader, Dataset
 
-def make_data_loaders(train_filename, val_filename, feature_set, num_workers, batch_size, filtered, random_fen_skipping, wld_filtered, skip_early_plies, param_index, main_device, epoch_size, val_size):
+def make_data_loaders(train_filename, val_filename, feature_set, num_workers, batch_size, filtered, random_fen_skipping, wld_filtered, early_fen_skipping, param_index, main_device, epoch_size, val_size):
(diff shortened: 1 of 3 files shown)


Sopel97:
I'm more inclined towards this generalized version. Will review tomorrow.

Sopel97:
Okay, I think the only issue right now is the parameter order is inconsistent. Either always put it before param index or after.

Eventually it's gonna need replacing with a struct because there's quite a lot of parameters right now, but for now at least let's not mess this up further.

Disservin:
Ah yeah the order was messed up also the parameter was missing in some places, using a struct is definitely a good idea, or even some json config file perhaps?

Sopel97:
Not mutually exclusive

train.py
```diff
@@ -57,6 +57,7 @@ def main():
   parser.add_argument("--epoch-size", type=int, default=100000000, dest='epoch_size', help="Number of positions per epoch.")
   parser.add_argument("--validation-size", type=int, default=1000000, dest='validation_size', help="Number of positions per validation step.")
   parser.add_argument("--param-index", type=int, default=0, dest='param_index', help="Indexing for parameter scans.")
+  parser.add_argument("--skip_early_plies", type=int, default=0, dest='skip_early_plies', help="Skip n plies from the start.")
```
linrock: naming nit: i'd suggest `--early-fen-skipping` for naming consistency with the other "skipping" flags (ie. `--no-smart-fen-skipping`, `--no-wld-fen-skipping`, `--random-fen-skipping`). otherwise, deferring to sopel and vondele for review

PR #223, opened 2023-01-01
Improve CUDA check for torch

vondele:


Sopel97:
I think a local import within `if pkg.is_version_at_least((1, 7)):` scope would be more sensible here. Also I think `from torch import cuda` should work and would import less things

vondele:
I do notice that (pre-existing) the version check is `if pkg.is_version_at_least((1, 7)):` but the error message is `LOGGER.error(f'Found torch version {pkg.version} but at least 1.8 required. Exiting.')`. Not sure what we require 1.7 or 1.8 ?

Sopel97:
Hmm, I'm not sure now. Though I think it should be okay if we raise the requirement to 1.8 now.

PR #227, opened 2023-01-23
Enable using early fen skipping param with easy_train.py

linrock:
Add support for the `--early-fen-skipping` param introduced in https://github.com/glinscott/nnue-pytorch/pull/222 to `easy_train.py`. This can be used to skip all positions where the ply count is less than or equal to the given number during training.


linrock:
fyi @vondele this was also used for training the new net, along with several recent master nets.

PR #228, opened 2023-02-17
Improve nnue.md documentation

TomAlard:
I've been using the `nnue.md` file to improve my own NNUE with quantization for some time now. It's been very helpful, but I've also found some issues, mostly with the optimized code. I've made a number of changes to the file which I split into independent commits, so I recommend reviewers to look at the individual commits (instead of all the changes simultaneously). Thank you for your time.

### DIFF
--- a/docs/nnue.md
+++ b/docs/nnue.md
@@ -1140,13 +1140,13 @@ int32_t* linear(
             m256_add_dpbusd_epi32(sum3, in, _mm256_load_si256(&layer.weights[offset3 + j * register_width]));
         }
 
-        const __m128i bias = _mm256_load_si256(&layer.bias[i * 4]);
+        const __m128i bias = _mm_load_si128(&layer.bias[i * 4]);
         // This function adds horizontally 8 values from each sum together, producing 4 int32 values.
         // For the definition see below.
         __m128i outval = m256_haddx4(sum0, sum1, sum2, sum3, bias);
         // Here we account for the weights scaling.
-        outval = _mm256_srai_epi32(outval, log2_weight_scale);
-        _mm256_store_si256(&output[i * 4], outval);
+        outval = _mm_srai_epi32(outval, log2_weight_scale);
+        _mm_store_si128(&output[i * 4], outval);
     }
 
     return output + layer.num_outputs;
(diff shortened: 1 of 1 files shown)


docs/nnue.md
```diff
@@ -1361,8 +1362,8 @@ int32_t* linear_sparse_input(
 
             m256_process_chunk(sum0, sum1, col0, _mm256_setzero_si256(), factor);
 
-            _mm256_store_si256(output + output_offset0, sum0);
-            _mm256_store_si256(output + output_offset1, sum1);
+            _mm256_store_si256(output + output_offset0, _mm256_srai_epi32(sum0, log2_weight_scale));
```
Sopel97: While there does indeed appear to be an omission regarding the output scaling, I believe this is wrong. Since this is inside `for (; i < num_nnz_input_indices; ++i) {` the same outputs can be hit multiple times - resulting in a recursive process similar to decaying average. Also, the scaling has to be applied to the main loop above too. Needs to be moved outside of these loops overall.

docs/nnue.md
```diff
@@ -1713,8 +1714,8 @@ int32_t* linear_sparse_input_block_sparse_output(
...
```
Sopel97: same as the first comment

Sopel97 (changes requested):
Thank you for fixing these errors. I believe there are a mistake introduced (same mistake in two place), however, which I detailed in the comments.

docs/nnue.md
TomAlard: Yes that makes sense. So how would you fix this? I was thinking of reverting this change and adding something like this to the end of the function
```c
for (int i = 0; i < layer.num_outputs; i += output_register_width) {
    _mm256_store_si256(output + i, _mm256_srai_epi32(_mm256_load_si256(output + i), log2_weight_scale));
}
```
Alternatively I could just revert these changes if you prefer. These two functions in particular are the only ones I haven't tried implementing myself, so I can't actually test this code.

Sopel97: That's the solution I had in mind. I think the change is important because otherwise the layer calculation is not complete.

TomAlard: Ok, I've added the changes.

Sopel97:
Looks good now, thanks.

PR #249, opened 2023-06-12
Add blocked sparse input affine transform approach to docs.

Sopel97:
Based on https://github.com/official-stockfish/Stockfish/pull/4612

### DIFF
--- a/docs/nnue.md
+++ b/docs/nnue.md
@@ -93,6 +93,8 @@ What this document DOES NOT contain:
             * [m256_haddx4](#m256_haddx4)
         - [Linear layer with sparse input](#linear-layer-with-sparse-input)
             * [m256_process_chunk](#m256_process_chunk)
+        - [Linear layer with blocked sparse input](#linear-layer-with-blocked-sparse-input)
+            * [Helping the odds](#helping-the-odds)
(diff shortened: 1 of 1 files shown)


vondele:
I have read through this, looks OK. 

Independent from this PR, should we integrate the permutation code (as mentioned https://github.com/official-stockfish/Stockfish/pull/4620#issue-1757202863) in the trainer somehow ?

Sopel97:
> Independent from this PR, should we integrate the permutation code (as mentioned https://github.com/official-stockfish/Stockfish/pull/4620#issue-1757202863) in the trainer somehow ?

We could maybe apply it during serialization if we have a good enough solution. Right now we should work on a separate tool. I made some initial scaffolding here https://github.com/Sopel97/nnue-pytorch/blob/ftperm/ftperm.py

vondele:
I think the greedy permutation that is used in the SF PR mentioned is likely a near-optimal solution to do this.

Computing the permutation needs some 50M positions, which in principle we can do by just using that many from a binpack and doing inference. Right now this is some modified version of SF collecting that data, but that's not ideal.

vondele:
anyway, docs seem complete. Ready for a merge?

Sopel97:
yes

PR #255, opened 2023-07-01
Increase L1 size to 2048 for training SFNNv7 nets

linrock:
This was used to train a larger master net (nn-1b951f8b449d.nnue)
https://github.com/official-stockfish/Stockfish/pull/4646

### DIFF
--- a/serialize.py
+++ b/serialize.py
@@ -297,7 +297,7 @@ def main():
   parser.add_argument("source", help="Source file (can be .ckpt, .pt or .nnue)")
   parser.add_argument("target", help="Target file (can be .pt or .nnue)")
   parser.add_argument("--description", default=None, type=str, dest='description', help="The description string to include in the network. Only works when serializing into a .nnue file.")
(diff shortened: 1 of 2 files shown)


serialize.py
```diff
@@ -318,7 +318,8 @@ def main():
     raise Exception('Invalid network input format.')
 
   if args.ft_compression != 'none' and not args.target.endswith('.nnue'):
-    raise Exception('Compression only allowed for .nnue target.')
+    args.ft_compression = 'none'
```
linrock, line 321: oops, i didn't mean to push this change to this PR. forgot there was an open PR with this branch. this was a hacky quick fix to get `--start-from-model` working when using leb128 as the default ft_compression arg. otherwise the trainer tries to serialize to .pt with leb128 compression and crashes upon start.

a cleaner fix may be to explicitly use `--ft_compression=none` when start models are prepared with `--start-from-model`

Sopel97: yes, I merged this because I realized my PR had a bug and this one was complete. I'm not sure what the best solution is.

PR #259, opened 2023-07-19
Add support for multiple datasets in the trainer. The datasets are in…

Sopel97:
…terleaved at binpack chunk granularity.

Handling cyclic binpack reading was moved from `BinpackSfenInputParallelStream` to `CompressedTrainingDataEntryParallelReader` to allow for cycling each dataset individually.

train.py accepts any number of positional arguments - paths to the datasets to use. Validation will use the same datasets unless overriden by `--validation-data`, which can be present multiple times (or have multiple values) to specify multiple datasets.

easy_train.py now can have multiple instances of `--training-dataset` and `--validation-dataset` (or accept multiple values for each).

C API changed, additional helper functions were made to wrap conversion of string list to array of char* for `create_fen_batch_stream` and `create_sparse_batch_stream`.

Also kinda deprecate .bin, currently won't work with multiple datasets. No one was using it anyway, we should remove it (at least don't allow it in the trainer) next time.

---------------------------------------------------------------------

I tested easy_train.py briefly, but I would like someone to do a full training run and compare the results with a manually interleaved dataset. The semantics should be similar, but there are some changes to the data loader that could manifest issues when re-reading datasets (requires a long training session), and also to make sure the interleaving is similar enough.

@linrock

### DIFF
--- a/train.py
+++ b/train.py
@@ -10,12 +10,12 @@ from torch import set_num_threads as t_set_num_threads
 from pytorch_lightning import loggers as pl_loggers
 from torch.utils.data import DataLoader, Dataset
 
-def make_data_loaders(train_filename, val_filename, feature_set, num_workers, batch_size, filtered, random_fen_skipping, wld_filtered, early_fen_skipping, param_index, main_device, epoch_size, val_size):
+def make_data_loaders(train_filenames, val_filenames, feature_set, num_workers, batch_size, filtered, random_fen_skipping, wld_filtered, early_fen_skipping, param_index, main_device, epoch_size, val_size):
   # Epoch and validation sizes are arbitrary
   features_name = feature_set.name
-  train_infinite = nnue_dataset.SparseBatchDataset(features_name, train_filename, batch_size, num_workers=num_workers,
+  train_infinite = nnue_dataset.SparseBatchDataset(features_name, train_filenames, batch_size, num_workers=num_workers,
                                                    filtered=filtered, random_fen_skipping=random_fen_skipping, wld_filtered=wld_filtered, early_fen_skipping=early_fen_skipping, param_index=param_index, device=main_device)
-  val_infinite = nnue_dataset.SparseBatchDataset(features_name, val_filename, batch_size, filtered=filtered,
+  val_infinite = nnue_dataset.SparseBatchDataset(features_name, val_filenames, batch_size, filtered=filtered,
                                                    random_fen_skipping=random_fen_skipping, wld_filtered=wld_filtered, early_fen_skipping=early_fen_skipping, param_index=param_index, device=main_device)
   # num_workers has to be 0 for sparse, and 1 for dense
   # it currently cannot work in parallel mode but it shouldn't need to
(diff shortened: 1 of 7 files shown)


training_data_loader.cpp
```diff
@@ -479,7 +479,7 @@ struct AsyncStream : Stream<StorageT>
 {
     using BaseType = Stream<StorageT>;
 
-    AsyncStream(int concurrency, const char* filename, bool cyclic, std::function<bool(const TrainingDataEntry&)> skipPredicate) :
+    AsyncStream(int concurrency, const std::vector<std::string>& filenames, bool cyclic, std::function<bool(const TrainingDataEntry&)> skipPredicate) :
         BaseType(1, filename, cyclic, skipPredicate)
```
linrock: got an error when building. changing `filename` to `filenames` here got compilation working

Sopel97: thanks. This constructor is not being ODR-used so MSVC didn't try compiling it.

Sopel97:
should be fixed now

train.py
```diff
@@ -61,10 +64,22 @@ def main():
   features.add_argparse_args(parser)
   args = parser.parse_args()
 
-  if not os.path.exists(args.train):
-    raise Exception('{0} does not exist'.format(args.train))
-  if not os.path.exists(args.val):
-    raise Exception('{0} does not exist'.format(args.val))
+  args.datasets = flatten_once(args.datasets)
+  if args.validation_datasets:
+    args.validation_datasets = flatten_once(args.validation_datasets)
```
linrock, line 69: crashing with another error now: 
```
File "/root/easy-train-data/experiments/experiment_L1-2048-new-117G/nnue-pytorch/train.py", line 176, in <module>
main()
File "/root/easy-train-data/experiments/experiment_L1-2048-new-117G/nnue-pytorch/train.py", line 78, in main
for val_dataset in args.validation_datasets:
TypeError: 'NoneType' object is not iterable
```

but works when setting validation_datasets to be the same as training_datasets:

```python
if args.validation_datasets:
  args.validation_datasets = flatten_once(args.validation_datasets)
else:
  args.validation_datasets = args.datasets
```

linrock:
how do you specify multiple training datasets with `easy_train.py`? if i use the same flag multiple times, only the last arg gets sent to `train.py`:

```
python3 easy_train.py \
  --training-dataset /data/UHO.e1fb1ade.binpack \
  --training-dataset /data/multinet_pv-2_diff-100_nodes-5000.binpack \
  --training-dataset /data/nodes5000pv2_UHO.binpack \
  ...
  ```
  
log output shows:
```
Training with: ['/data/nodes5000pv2_UHO.binpack', '/data/nodes5000pv2_UHO.binpack']
Validating with: ['/data/nodes5000pv2_UHO.binpack', '/data/nodes5000pv2_UHO.binpack']
```

File: train.py (hunk 2 above)
Comment (Sopel97) [reply train.py:69]: hmm, it looks like argparse sets it to None instead of an empty array, which is pretty dumb but ok. Should be fixed now, I hope...

**Comment by linrock** [comment]
oops my bad, i forgot to update the branch from which i was invoking `easy_train.py`. i'm seeing the same results as you now. looks to be working as expected!

**Comment by linrock** [comment]
i started training runs with both: a manually interleaved binpack + using multiple training datasets as runtime flags. i'll report back later with how they compare

**Comment by linrock** [comment]
current status of trainings from scratch:
- L1-2048-new-117G - (blue) 2x runs with manually interleaved binpacks using interleave_binpacks.py
- L1-2048-new-117G-individual- (orange) 2x runs using this branch and several --training-dataset flags
 
image


117G total size of binpacks of different sizes, all compressed:

```
 38G nodes5000pv2_UHO.binpack
 26G multinet_pv-2_diff-100_nodes-5000.binpack
 21G UHO.e1fb1ade.binpack
9.7G dfrc-n5000.L1-2048.binpack
9.2G UHO_4060_v2.1.binpack
8.2G test60-2019-2tb7p.binpack
3.1G UHO.L1-2048.binpack
3.1G UHO-4060-v2.L1-2048.binpack
```

default LR and gamma. lambda 1.0 for the first 500 epochs, then end-lambda set to 0.75 at epoch 500.

Sopel97:
I did a test with `farseerT74.binpack`, `farseerT75.binpack`, `farseerT76.binpack`. Results:

plot_440_441

440 - interleaved in the data loader
441 - interleaved by interleave_binpacks

It looks like they are close to equivalent. I will, however remove the requirement for `filesystem` because it's unnecessary and causes issues on older gcc.

vondele:
This looks ready to merge, OK?

Sopel97:
yes, tests seem to indicate it works correctly

PR #264, opened 2023-09-14
Add complete requirements.txt. Update readme on the installation process.

Sopel97:
I can split this into PRs if deemed so. The readme cleanup was kinda necessary in this scope because it was very disjoint and this patch touches the whole setup process.

I tested this on a clear venv with Python 3.11.5 on CentOS 8, running A100s.

CUDA requirement might need adjusting as time goes on and new GPUs are released, but it should at least be backwards compatible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,30 +1,44 @@
 # Setup
 
-#### Install PyTorch
+### Software prerequisites
 
-[PyTorch installation guide](https://pytorch.org/get-started/locally/)
+#### Python
+
+If using easy_train.py then at least 3.7 is required.
+
+Otherwise versions around 3.6 should also work, but updating is recommended.
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,30 +1,44 @@
...
-```
+If using easy_train.py then at least 3.7 is required.
+
+Otherwise versions around 3.6 should also work, but updating is recommended.
+
+Python 3.11 was also tested and works well.
+
+#### A C++ compiler
+
+If you're using easy_train.pt then at least GCC 9.2 is required for compiling the data loader, Stockfish, and c-chess-cli. This a defensive version requirement as older versions were known to cause problems with Stockfish.
```
Disservin: .py instead of .pt ^^

PR #267, opened 2023-10-04
Add SFNNv8 architecture and diagrams

linrock:
L1-2560 introduced in https://github.com/official-stockfish/Stockfish/pull/4795


dav1312:
For SFNNv7 and v8 what about saying `Same as "SFNNv5"` to lead directly to the original arch?
Instead of Same as "SFNNv7" > Same as "SFNNv6" > Same as "SFNNv5"

linrock:
> is the v1 needed anywhere?

regarding the diagrams? i only see "detailed_v2" diagrams used in this repo. not sure if others are used elsewhere. all the architecture diagrams currently have 3 versions.

> For SFNNv7 and v8 what about saying Same as "SFNNv5" to lead directly to the original arch?

descriptions updated to refer to SFNNv5

PR #268, opened 2023-10-24
Update default testing book to UHO_Lichess_4852_v1.epd.zip

linrock:
To match the recent new fishtest book


vondele:
LGTM, the difference will be very small though.

PR #277, opened 2024-03-15
Add formatting

Disservin:
@Sopel97 what do you say about adding some formatters? 

black is also used for the wdl repo https://github.com/official-stockfish/WDL_model/tree/master, and the clang-format is taken from stockfish master

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,44 @@
+AccessModifierOffset: -1
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: Consecutive
+AlignConsecutiveDeclarations: Consecutive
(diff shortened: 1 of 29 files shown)


Sopel97:
if we can get it to agree whether strings should be `''` or `""` I'm fine with it

Disservin:
It uses double quotes `"` but some strings have double quotes inside them and use the opposite quotation characters. I.e. 
`print('root_dir - the directory to "cleanup"')`, this isn't automatically converted into a double quote because it'd need escaping.

vondele:
I would also add a formatter, so if this can be rebased, let's merge.

PR #278, opened 2024-03-30
Fix Typo

xu-shawn:
This fixes a typo present in both SFNNv8 images

See:
https://discord.com/channels/435943710472011776/718853716266188890/1222034111930044446
https://discord.com/channels/435943710472011776/718853716266188890/1218367741749362790

PR #280, opened 2024-04-30
Increase L1 size to 3072

linrock:
To match SFNNv9: https://github.com/official-stockfish/Stockfish/pull/5149


cj5716:
I think we can incorporate this https://github.com/official-stockfish/nnue-pytorch/pull/278 PR here as well

PR #282, opened 2024-05-30
use pdep to find nthSetBitIndex

Disservin:
a bit better than the current approach, current approach shows up in profile with ~10% overhead
https://godbolt.org/z/avfT1jY3Y

cmake checks for zen3+ cpu


Sopel97:
Good to see this change, completely forgot about it.

Should work in MSYS too from the looks of it, but I'm concerned about plain windows. Can we make sure it at least compiles? I.e. run the check only on linux

Sopel97 (approved):
looks good and seems to compile in msys

PR #284, opened 2024-06-19
Add SFNNv9 Diagram

xu-shawn:


Disservin:
didn't look too carefully but LGTM

PR #296, opened 2025-04-05
Add requests to requirements.txt

simonstratmann:
requests is required


Disservin:
thanks requests was a part of https://github.com/official-stockfish/nnue-pytorch/pull/299 and the cmake change shouldn't be required anymore I think

PR #309, opened 2025-06-15
Enhance dataloader benchmark

vondele:


training_data_loader.cpp
```diff
@@ -1079,31 +1079,72 @@ extern "C" {
...
-//     auto t1 = std::chrono::high_resolution_clock::now();
-
-//     std::cout << (t1 - t0).count() / 1e9 << "s\n";
-//     return 0;
-// }
+#if defined(BENCH)
+
+/* benches
+   compile and run with:
+     g++ -g3 -O3 -DNDEBUG -DBENCH -march=native training_data_loader.cpp && ./a.out /path/to/binpack
```
Disservin, line 1086: why DBENCH ?

training_data_loader.cpp
```diff
@@ -1079,31 +1079,72 @@ extern "C" {
...
+{
+    if (argc < 2) {
+        std::cerr << "Usage: " << argv[0] << " file1 [file2 ...]\n";
+        return 1;
+    }
+
+    const char** files = const_cast<const char**>(&argv[1]);
+    int file_count = argc - 1;
+
+    const int concurrency = 64;
```
Disservin: maybe `std::min(64, std::thread::hardware_concurrency())` ?

training_data_loader.cpp
vondele: Benches was the comment before, so I used that for the define. Other suggestions welcome

training_data_loader.cpp
vondele: Even without the min... I agree

training_data_loader.cpp
Disservin: yeah but I dont see the define being used anywhere or am I blind?

vondele: 3 lines above?
`#if defined(BENCH)`

Disservin: oh wow i'm blind :D thanks

Issue #298, opened 2025-04-18
Trainer does not work with read-only binpacks


protonspring:
I made some binpacks and made them read-only so they couldn't ever be accidentally changed or deleted.

If I try to use them in the nnue-pytorch trainer it just hangs. . no errors or anything.

When I change the currently used .binpack to RW, then everything works fine.

Is that intended?  Why would the trainer want to change anything in the .binpack?

Sopel97:
I don't remember exactly, but I think it's always opened in append mode for simplicity.

protonspring:
Would you like me to make a PR, or maybe no one cares. 

I think the trainer should at least tell the user what's wrong.

Disservin:
problem is the compressedtrainingdatafile I think, 
https://github.com/official-stockfish/nnue-pytorch/blob/42abd2fa59bbacc7b57d0d97a81d4502e9eb4f6c/lib/nnue_training_data_formats.h#L6752 which handles reading and writing so it opens them in both

PR #312, opened 2025-06-15
Allow for readonly binpacks

Disservin:
fixes https://github.com/official-stockfish/nnue-pytorch/issues/298


vondele:
nice, that should allow a /workspace/data to be mounted read-only in a container that does only training.

PR #314, opened 2025-06-16
AMD Docker Support

xu-shawn:


Dockerfile.AMD
```diff
@@ -0,0 +1,19 @@
...
+    git \
+    curl \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /workspace/nnue-pytorch
+
+COPY requirements.txt setup_script.sh ./
+
+RUN pip install --no-cache-dir -r requirements.txt
+RUN ROCM_HOME=/opt/rocm CUPY_INSTALL_USE_HIP=1 pip install --no-cache-dir git+https://github.com/cupy/cupy.git
```
Disservin: Can this be pinned to a SHA?

xu-shawn: done

PR #316, opened 2025-06-17
Make use_cupy an argument, and default true

vondele:


Disservin:
changed a few things
```
show help when no Arguments
fix 3 parameters in eval_perm_impl call
rename option to "--no-cupy" instead of "--use-cupy"
add cupy device parameter
```

PR #319, opened 2025-06-20
Prefetching

vondele:


nnue_dataset.py
```diff
@@ -496,15 +496,68 @@ def __iter__(self):
...
+
     def __len__(self):
         return self.num_batches
 
     def __getitem__(self, idx):
-        return next(self.iter)
+        self._start_prefetching()
+
+        try:
+            item = self._prefetch_queue.get(timeout=30.0)  # 30 second timeout
```
Sopel97, line 547: I'm pretty sure this would make the trainer unsafe for suspension, though I'm not sure if it ever was.

nnue_dataset.py
```diff
@@ -496,15 +496,68 @@ def __iter__(self):
...
+
+
 class FixedNumBatchesDataset(Dataset):
     def __init__(self, dataset, num_batches):
         super(FixedNumBatchesDataset, self).__init__()
         self.dataset = dataset
         self.iter = iter(self.dataset)
         self.num_batches = num_batches
 
+        self._prefetch_queue = queue.Queue(maxsize=100)
```
Sopel97, line 511: for batch size 16384 one batch should be around 16MB, if I'm counting properly, so this *should* be fine, but I'm concerned about high worker numbers. Warrants testing in a scenario where the queues get full.

vondele: I have tested with 64 worker threads, which worked fine for me. Not sure if 100 was picked just like that or if there would be some need to tune it. Given that training now requires quite high-end hardware to be practical, I think we can work with the assumption of fairly large memory.

nnue_dataset.py
vondele: I dont think suspending is the use case, but maybe we need to be a bit more generous nevertheless.

nnue_dataset.py
Disservin: I think I just picked 100 like that but even locally used like 500

vondele: so, let's go with the current 100, and see later on. I think the real trick will be to get to more overlap.

PR #321, opened 2025-06-24
Implement simple_eval_skipping

vondele:
as introduced by @linrock for training small nets, see e.g. https://github.com/official-stockfish/Stockfish/commit/cb4a62311985f685ba6f5457851527a3289073e6

https://github.com/linrock/nnue-pytorch/tree/128-skipSimpleEval-lt950-torch-compile


Sopel97:
I think it would be less confusing if simple_eval returned a correctly signed evaluation

PR #326, opened 2025-06-28
Abstract Dataloader Skip Config

Disservin:
Introduce a struct for this mess.
Implements https://github.com/official-stockfish/nnue-pytorch/pull/222#issuecomment-1368183732


vondele:
Thanks!

Disservin:
Just saw that I didn't update the bench part of the dataloader will fix that in a separate pr..

PR #333, opened 2025-07-14
Introduce a dataclass for loss parameters

vondele:
Some cleanup, and use it to expose the params on the commandline.

### DIFF
--- a/model.py
+++ b/model.py
@@ -11,7 +11,7 @@ L2 = 15
 L3 = 32
 
 
-# parameters needed for the definition of hte loss
+# parameters needed for the definition of the loss
 @dataclass
 class LossParams:
     in_offset: float = 270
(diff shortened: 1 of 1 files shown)


model.py
```diff
@@ -3,13 +3,47 @@
...
+        qp_asymmetry: float = 0.0,
+    ):
+        self.in_offset = in_offset
+        self.out_offset = out_offset
+        self.in_scaling = in_scaling
+        self.out_scaling = out_scaling
+        self.start_lambda = start_lambda
+        self.end_lambda = end_lambda
+        self.pow_exp = pow_exp
+        self.qp_asymmetry = qp_asymmetry
```
Disservin: This bit gets auto generated with the dataclass tag

PR #338, opened 2025-09-18
Compile dataloader with pgo

xu-shawn:
Seems like a speedup locally
```
❯ g++ -std=c++20 -g3 -O3 -DNDEBUG -DBENCH -march=native training_data_loader.cpp
❯ ./a.out t77_78_79/test77-dec2021-16tb7p.no-db.min.binpack
Iter:       6000       Time(s):   79.311       MPos/s:      1.239       It/s:        75.7       MB/s:       233.2       B/pos:       197
❯ g++ -std=c++20 -g3 -O3 -DNDEBUG -DBENCH -march=native -fprofile-use=./build-pgo-generate/pgo_data -fprofile-correction training_data_loader.cpp
training_data_loader.cpp: In function ‘(static initializers for training_data_loader.cpp)’:
training_data_loader.cpp:1159:1: warning: ‘./build-pgo-generate/pgo_data/#home#shawn#Documents#Development#Python#nnue-pytorch#a-training_data_loader.gcda’ profile count data file not found [-Wmissing-profile]
 1159 | }
      | ^
❯ ./a.out t77_78_79/test77-dec2021-16tb7p.no-db.min.binpack
Iter:       6000       Time(s):   75.869       MPos/s:      1.296       It/s:        79.1       MB/s:       244.3       B/pos:       197
```


compile_data_loader.bat
```diff
@@ -1,2 +1,10 @@
-cmake . -Bbuild -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX="./"
-cmake --build ./build --config RelWithDebInfo --target install
\ No newline at end of file
+cmake -S . -B build-pgo-generate -DCMAKE_BUILD_TYPE=PGO_Generate
+cmake --build ./build-pgo-generate --config PGO_Generate
+
+./build-pgo-generate/training_data_loader_benchmark /home/shawn/Documents/Development/Python/nnue-pytorch/test77-dec2021-16tb7p.no-db.min.binpack
```
vondele: should be possible without hard-coded binpack path

vondele: maybe as an argument to the script?

xu-shawn: Since the script is run at docker entry, I'm wondering if it's better to directly include a tiny binpack in the repository?

vondele: also an option.

PR #348, opened 2025-09-21
improve types from data_loader

Disservin:


vondele:
works for me
```
Epoch 3: 100%|██████████| 6104/6104 [00:36<00:00, 166.90it/s, v_num=0]`Trainer.fit` stopped: `max_epochs=4` reached.

```

PR #349, opened 2025-09-23
Modularize NNUE

xu-shawn:


train.py
```diff
@@ -135,7 +136,7 @@ def main():
         default="30:00:00:00",
         type=str,
         dest="max_time",
-        help="The maximum time to train for. A string in the format DD:HH:MM:SS (Default 30:00:00:00).",
+        help="The maximum time to train for. A string in the format DD:H:MM:SS (Default 30:00:00:00).",
```
vondele: that doc change is probably wrong?

xu-shawn: thanks for the catch. somehow made a typo :facepalm:

model/model.py
```diff
@@ -1,46 +1,29 @@
...
+from .config import ModelConfig
+from .feature_transformer import DoubleFeatureTransformerSlice
 
 
 class LayerStacks(nn.Module):
-    def __init__(self, count: int):
+    def __init__(self, count: int, config: ModelConfig = ModelConfig()):
         super(LayerStacks, self).__init__()
 
+        self.config = config
```
Disservin: don't think we need to persist self.config here?

model/model.py
```diff
@@ -139,18 +124,25 @@ class NNUEModel(nn.Module):
     def __init__(
         self,
         feature_set: FeatureSet,
+        config: ModelConfig = ModelConfig(),
         num_psqt_buckets: int = 8,
         num_ls_buckets: int = 8,
     ):
         super().__init__()
+
+        self.config = config
```
Disservin: same

model/model.py
```diff
@@ -139,18 +124,25 @@ class NNUEModel(nn.Module):
     def __init__(
         self,
         feature_set: FeatureSet,
+        config: ModelConfig = ModelConfig(),
```
Disservin: this is prone to errors imo because the model config should be cli configurable so should be a required argument and not default to something, same earlier

PR #352, opened 2025-09-24
ddp loader

Disservin:


Disservin:
@Sopel97 do you think this would work regarding the changes for the multi gpu training ?

lib/nnue_training_data_formats.h
```diff
@@ -6794,6 +6794,49 @@ namespace binpack
...
+            if (world_size <= 1) {
+                seek_to_start();
+                return;
+            }
+            
+            // Calculate the target position for this rank
+            std::size_t target_position = (m_sizeBytes * rank) / world_size;
+            
+            // Start from the beginning of the file
+            seek_to_start();
```
Sopel97:
```
            seek_to_start();

            if (world_size <= 1)
                return;

            // Calculate the target position for this rank
            std::size_t target_position = (m_sizeBytes * rank) / world_size;
```

lib/nnue_training_data_formats.h
```diff
@@ -7816,6 +7866,11 @@ namespace binpack
         std::function<bool(const TrainingDataEntry&)> m_skipPredicate;
 
         std::vector<std::thread> m_workers;
+        
+        // DDP support
+        int m_rank;
+        int m_world_size;
+        std::vector<bool> m_files_seeked_for_ddp;  // Track which files have been seeked for DDP
```
Sopel97: I'd replace with uint8_t. Really no need to bother with `vector<bool>` and all its potential problems

lib/nnue_training_data_formats.h
```diff
@@ -7882,25 +7872,36 @@ namespace binpack
...
                     else
                         return true;
                 }
 
                 m_chunk = inputFile.readNextChunk();
                 m_offset = 0;
+
+                if (m_world_size > 1)
+                {
+                    m_ddp_chunks_to_skip_after_read[fileId] = (m_world_size > 0 ? static_cast<std::size_t>(m_world_size - 1) : 0);
```
Sopel97: pretty sure `m_ddp_chunks_to_skip_after_read` is size() == 0 here.

Disservin:
> but at most once to prevent infinite loop

but we are using this m_cyclic with true ? so we can just seek to start, skip chunks and be done with it, not sure which at most once logic you are refering to?

Disservin:
Yeah it's an artifact just wanted to clear things up

vondele:
I think the initialization trick with the environment variables is not yet really working, at least locally I get this:
```
DDP rank: 0, world size: 1
DDP rank: 0, world size: 1
Initializing distributed: GLOBAL_RANK: 0, MEMBER: 1/4
DDP rank: 2, world size: 4
DDP rank: 3, world size: 4
DDP rank: 2, world size: 4
DDP rank: 1, world size: 4
DDP rank: 3, world size: 4
DDP rank: 1, world size: 4
Initializing distributed: GLOBAL_RANK: 3, MEMBER: 4/4
Initializing distributed: GLOBAL_RANK: 2, MEMBER: 3/4
Initializing distributed: GLOBAL_RANK: 1, MEMBER: 2/4
distributed_backend=nccl
All distributed processes registered. Starting with 4 processes
```
So rank 0 thinks the world size is 1 at the time of creating the stream?

vondele:
I'm looking at the training loss, and I think we're not yet fully equivalent, consistent with the result from training (but looking at the validation loss might be quicker).

image

PR #372, opened 2025-11-11
Simplify double feature transformer slice

xu-shawn:
not sure if this will be a slowdown


vondele:
```
master
------
75.82it/s, 76.27it/s

simplify_ft
-----------
73.64it/s, 74.43it/s
```
Looks like the same speed.

PR #373, opened 2025-11-26
Add Threat Input support

sscg13:
This is the corresponding PR to https://github.com/official-stockfish/Stockfish/pull/6411 and the training code for the latest SF master networks.

Special thanks to @jw1912 whose implementation of threat inputs in bullet is almost verbatim translated here.

### DIFF
--- a/docs/nnue.md
+++ b/docs/nnue.md
@@ -2987,7 +2987,7 @@ y = self.layer_stacks(l0_, layer_stack_indices) + (wpsqt - bpsqt) * (us - 0.5)
 
 ### "SFNNv10" architecture
 
-Same as "SFNNv5" with Full_Threat input features added.
+Same as "SFNNv5" with Full_Threat input features added and the feature transformer quantized one changed to 255.
 
 2025-11-12 - *
 


model/quantize.py
```diff
@@ -63,14 +63,16 @@ def quantize_feature_transformer(
...
-        weight = weight.mul(self.quantized_one).round().to(torch.int16)
+        bias = bias.mul(self.ft_quantized_one).round().to(torch.int16)
+        weight = weight.mul(self.ft_quantized_one).round().to(torch.int16)
         psqt_weight = (
             psqt_weight.mul(self.nnue2score * self.weight_scale_out)
             .round()
             .to(torch.int32)
         )
 
+        weight[0:79856] = weight[0:79856].clamp(-128, 127)
```
PikaCat-OuO: The weights are still getting clamped here with non-threat feature set.

sscg13: thanks for pointing that out!

model/model.py
```diff
@@ -116,6 +116,15 @@ def clip_weights(self):
                             raise Exception("Not supported.")
                     p.data.copy_(p_data_fp32)
 
+    def clip_threat_weights(self):
+        if self.feature_set.name[:12] == "Full_Threats":
```
Disservin: could also be 

.startswith()

but nitpicking

sscg13: ok

sscg13:
how else should we make the indexing static then?

Disservin:
basically just make `init_threat_offsets` constexprable and we can have `threatoffsets` we only need to change some attack generation to be constexprable and the bb counting, i'll push something

PR #375, opened 2025-12-02
Backport changes needed for master nets

vondele:
this introduces two changes needed for training SF master nets, in particular for executing the recipes referenced in

https://github.com/official-stockfish/Stockfish/pull/6452

https://github.com/official-stockfish/Stockfish/pull/6457

adding some additional flexibility for shaping the piece count distribution and weighting the individual configurations in the loss respectively.

### DIFF
--- a/training_data_loader.cpp
+++ b/training_data_loader.cpp
@@ -1029,12 +1029,13 @@ struct FenBatchStream: Stream<FenBatch> {
 };
 
 struct DataloaderSkipConfig {
-    bool filtered;
-    int  random_fen_skipping;
-    bool wld_filtered;
-    int  early_fen_skipping;
-    int  simple_eval_skipping;
-    int  param_index;
+    bool   filtered;
+    int    random_fen_skipping;
+    bool   wld_filtered;
+    int    early_fen_skipping;
+    int    simple_eval_skipping;
+    int    param_index;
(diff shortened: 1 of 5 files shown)


training_data_loader.cpp
```diff
@@ -1049,19 +1050,17 @@ std::function<bool(const TrainingDataEntry&)> make_skip_predicate(DataloaderSkip
...
+            auto desired_piece_count_weights = [&config](int pc) -> double {
+                double x  = pc;
+                double x1 = 0, y1 = config.pc_y1;
+                double x2 = 16, y2 = config.pc_y2;
+                double x3 = 32, y3 = config.pc_y3;
+                double l1 = (x - x2) * (x - x3) / ((x1 - x2) * (x1 - x3));
+                double l2 = (x - x1) * (x - x3) / ((x2 - x1) * (x2 - x3));
+                double l3 = (x - x1) * (x - x2) / ((x3 - x1) * (x3 - x2));
+                return l1 * y1 + l2 * y2 + l3 * y3;
+            };
```
Sopel97, line 1063: if the filtering performance becomes an issue these could be computed in the capture

vondele: performance so far is OK, but I agree. It would be nice if they are always consistent with the parameters y1..y3 though.

PR #376, opened 2025-12-02
Rename feature transformer and kernels

xu-shawn:


model/features/halfka_v2_hm.py
```diff
@@ -2,6 +2,7 @@
 
 import chess
 
+from ..modules import DoubleFeatureTransformer
```
Disservin: why is this needed?

xu-shawn: accidential change oops

PR #383, opened 2025-12-19
Print threat weights directly as i8.

sscg13:
Companion PR to https://github.com/official-stockfish/Stockfish/pull/6479

### DIFF
--- a/model/utils/serialize.py
+++ b/model/utils/serialize.py
@@ -319,7 +319,7 @@ class NNUEReader:
         if self.feature_set.name.startswith("Full_Threats"):
             threat_weight = self.tensor(np.int8, [79856, shape[1] - num_psqt_buckets])
             psq_weight = self.tensor(np.int16, [shape[0] - 79856, shape[1] - num_psqt_buckets])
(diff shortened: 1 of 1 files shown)


model/utils/serialize.py
```diff
@@ -310,7 +316,12 @@ def read_feature_transformer(
 
         bias = self.tensor(np.int16, [layer.bias.shape[0] - num_psqt_buckets])
         # weights stored as [num_features][outputs]
-        weight = self.tensor(np.int16, [shape[0], shape[1] - num_psqt_buckets])
+        if self.feature_set.name.startswith("Full_Threats"):
+            threat_weight = self.tensor(np.int8, [79856, shape[1] - num_psqt_buckets])
+            psq_weight = self.tensor(np.int16, [shape[0] - 79856, shape[1] - num_psqt_buckets])
+            weight = torch.cat(threat_weight, psq_weight, dim=0)
```
Disservin: i think you need an array for torch.cat?
`torch.cat([threat_weight, psq_weight], dim=0)`

sscg13: ah oops

PR #386, opened 2026-02-01
chore: pin pytorch version to 2.10.0

AlpinDale:
The torch version shouldn't be loose. 

Tested with the latest stable torch (2.10.0) and training works fine.


Disservin:
we are basically only supporting the container way to run this and the nvidia pytorch image already includes a pinned torch version.

PR #391, opened 2026-02-18
run_docker.sh fix

TonyCongqianWang:
It is a good addition to let people run docker without root rights (e.g. on some cluster).

Right now when trying to run in rootless mode, python will throw an Exception due to an unknown uid. When run in rootless mode, the only valid uid becomes 0.

With this fix run_docker.sh automatically detects if docker is configured in rootless mode and will pass the correct user flag to the docker run command.

### DIFF
--- a/run_docker.sh
+++ b/run_docker.sh
@@ -33,10 +33,19 @@ read DATA_PATH
 DATA_PATH=${DATA_PATH}
 echo "Using data path: $DATA_PATH"
 
+# Checking if docker is in rootless mode.
+if docker info 2>/dev/null | grep -iq "rootless"; then
+    echo "Rootless mode detected."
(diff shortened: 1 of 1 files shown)


Disservin:
i dont really understand this, `id -u` is intended to match the user id that you are running the script from.. so that new files have the correct permissions, now if you have docker rootless and run it, you're gonna endup with a root owned files which don't match the hosts id ?

TonyCongqianWang:
Honestly I don't understand it anymore either.

This fix worked for me but I can't tell you why, maybe some else should either explain why it works or come up with a different fix.

Without this the training script refused to run, because for some reason the only detected user in the rootless docker was 0.

Anyway the permissions aren't an issue because you run it as rootless, you don't actually have root permissions and all files belong to you

PR #392, opened 2026-02-18
Fix: Remove some unnecessary tensor copy operations from model

TonyCongqianWang:
Right now there are some unnecessary tensor copies particularly for weight clipping. Omitting those copies doesn't seem t speed up the training, but it is arguably cleaner

### DIFF
--- a/model/model.py
+++ b/model/model.py
@@ -1,4 +1,5 @@
 import torch
+import torch.nn.functional as F
 from torch import nn
 
 from .config import ModelConfig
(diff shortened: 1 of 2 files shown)


model/model.py
```diff
@@ -99,32 +95,23 @@ def clip_weights(self):
...
+                            max_weight = (
                                 p_data_fp32.new_full(p_data_fp32.shape, max_weight)
                                 - expanded_virtual_layer
                             )
-                            p_data_fp32 = torch.min(p_data_fp32, max_weight_t)
-                    else:
-                        if min_weight is not None and max_weight is not None:
-                            p_data_fp32.clamp_(min_weight, max_weight)
-                        else:
-                            raise Exception("Not supported.")
```
TonyCongqianWang, line 117: I removed the Exception. Not sure if there is any reason to keep it. To me at least it makes no logical sense to allow None for min AND max but not for either one alone

PR #405, opened 2026-02-24
Use Tyro for CLI Args

Disservin:
only for train.py so far 

renamed -no-smart-fen-skipping and --no-wld-fen-skipping
```
 --filtered, --no-filtered                                                                                                                                                                           
                      If disabled, no smart fen skipping will be done. (default: True)                                                                                                               
 --wld-filtered, --no-wld-filtered                                                                                                                                                                   
                      If disabled, WLD-based fen skipping is turned off. (default: True)       
```

### DIFF
--- a/data_loader/config.py
+++ b/data_loader/config.py
@@ -15,13 +15,13 @@ class DataloaderSkipConfig:
     simple_eval_skipping: int = -1
     """Skip positions based on simple eval. -1 = disabled."""
     param_index: int = 0
-    """Parameter index for piecewise cubic scaling."""
+    """Indexing for parameter scans."""
     pc_y1: float = 1.0
-    """Piecewise cubic y1 parameter."""
+    """Piecewise quadratic interpolation y1 parameter."""
     pc_y2: float = 2.0
-    """Piecewise cubic y2 parameter."""
+    """Piecewise quadratic interpolation y2 parameter."""
     pc_y3: float = 1.0
-    """Piecewise cubic y3 parameter."""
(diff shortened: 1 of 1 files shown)


data_loader/config.py
```diff
@@ -4,92 +4,31 @@
...
-    wld_filtered: bool = False
+    """Skip a random fraction of positions. 0 = disabled."""
     early_fen_skipping: int = -1
+    """Skip positions from the start of the game. -1 = disabled."""
     simple_eval_skipping: int = -1
+    """Skip positions based on simple eval. -1 = disabled."""
     param_index: int = 0
+    """Parameter index for piecewise cubic scaling."""
     pc_y1: float = 1.0
+    """Piecewise cubic y1 parameter."""
```
vondele: Piece count, cubic interpolation y1 parameter

data_loader/config.py
```diff
@@ -4,92 +4,31 @@
...
-            help="piece count parameter y2 (default=2.0)",
-        )
-        parser.add_argument(
-            "--pc-y3",
-            type=float,
-            default=1.0,
-            dest="pc_y3",
-            help="piece count parameter y3 (default=1.0)",
-        )
+    """Piecewise cubic y3 parameter."""
```
vondele: same for y2, y3

data_loader/config.py
vondele: actually quadratic

Disservin:
tested fine for me now

PR #407, opened 2026-02-26
Improving dataloader cpp and compile script

TonyCongqianWang:
- compile file was misleading because it did not use correct .bat syntax -> replaced with cross plattform python script
- compile used pgo, but pgo did not work correctly in docker containers. Removed for now as docker is the main usage
- optimized some inefficient filtering logic by precalulating lookup tables
- added functionality to soft filter early plies


vondele:
pgo was added recently, so better no remove it, rather fix. It is a measurable speedup.

TonyCongqianWang:
Closing in favour of #430

PR #413, opened 2026-03-01
Improving data_loader file structure and usability

TonyCongqianWang:
From the last PR 2 more cpp related files were created. This clutters the root directory. As such I propose to move them to data_loader/cpp. I additionally added some command line args that improve the usability of the benchmarking tool as well as guarding against broken file paths that lead to segfaults otherwise.

It seems that the pgo benchmark we have been using has been cheating since it was compiled as a monolith, the abi barriers are not respected and as such the performance is not indicative of real use. However it seems that pgo still consistently outperforms non-pgo when using the shared library:

```
root@b6ac1a527fe6:/workspace/nnue-pytorch# ./bench_shared .pgo/small.binpack 
Iter:     6000   Time(s):     9.23   MPos/s:     10.7   It/s:    7e+02   MB/s:    3e+03   B/pos:    3e+02
root@b6ac1a527fe6:/workspace/nnue-pytorch# ./bench_shared .pgo/small.binpack 
Iter:     6000   Time(s):     9.29   MPos/s:     10.6   It/s:    6e+02   MB/s:    3e+03   B/pos:    3e+02
root@b6ac1a527fe6:/workspace/nnue-pytorch# g++ -std=c++20 -g3 -O3 -DNDEBUG -march=native -fPIC -shared \
    data_loader/cpp/training_data_loader.cpp \
    -o build/libtraining_data_loader.so
root@b6ac1a527fe6:/workspace/nnue-pytorch# ./bench_shared .pgo/small.binpack 
Iter:     6000   Time(s):     9.81   MPos/s:       10   It/s:    6e+02   MB/s:    3e+03   B/pos:    3e+02
root@b6ac1a527fe6:/workspace/nnue-pytorch# ./bench_shared .pgo/small.binpack 
Iter:     6000   Time(s):     9.82   MPos/s:       10   It/s:    6e+02   MB/s:    3e+03   B/pos:    3e+02
root@b6ac1a527fe6:/workspace/nnue-pytorch# 
```

Note that the benchmark compiled by the cmake is instrumented and extremely slow. It might be better to produce two different executables as this could be misleading.

```
root@b6ac1a527fe6:/workspace/nnue-pytorch# ./build/training_data_loader_bench ./.pgo/small.binpack 
Iter:       30   Time(s):     11.1   MPos/s:   0.0444   It/s:        3   MB/s:    2e+01   B/pos:    5e+02
```

Library speed seems similar between master and patch. I can't measure accurately enough, but there could be a very slight degradation for whatever reason, even though it is more likely just noise.

MASTER
```
Iter:     6000   Time(s):     9.07   MPos/s:     10.8   It/s:    7e+02   MB/s:    3e+03   B/pos:    3e+02
Iter:     6000   Time(s):     9.08   MPos/s:     10.8   It/s:    7e+02   MB/s:    3e+03   B/pos:    3e+02
Iter:     6000   Time(s):     9.07   MPos/s:     10.8   It/s:    7e+02   MB/s:    3e+03   B/pos:    3e+02
Iter:     6000   Time(s):     9.01   MPos/s:     10.9   It/s:    7e+02   MB/s:    3e+03   B/pos:    3e+02
Iter:     6000   Time(s):     9.06   MPos/s:     10.9   It/s:    7e+02   MB/s:    3e+03   B/pos:    3e+02
Iter:     6000   Time(s):      9.1   MPos/s:     10.8   It/s:    7e+02   MB/s:    3e+03   B/pos:    3e+02
Iter:     6000   Time(s):     9.05   MPos/s:     10.9   It/s:    7e+02   MB/s:    3e+03   B/pos:    3e+02

```

PATCH
```
Iter:     6000   Time(s):     9.05   MPos/s:     10.9   It/s:    7e+02   MB/s:    3e+03   B/pos:    3e+02
Iter:     6000   Time(s):     9.31   MPos/s:     10.6   It/s:    6e+02   MB/s:    3e+03   B/pos:    3e+02
Iter:     6000   Time(s):      9.1   MPos/s:     10.8   It/s:    7e+02   MB/s:    3e+03   B/pos:    3e+02
Iter:     6000   Time(s):      9.1   MPos/s:     10.8   It/s:    7e+02   MB/s:    3e+03   B/pos:    3e+02
Iter:     6000   Time(s):     9.03   MPos/s:     10.9   It/s:    7e+02   MB/s:    3e+03   B/pos:    3e+02
Iter:     6000   Time(s):     9.28   MPos/s:     10.6   It/s:    6e+02   MB/s:    3e+03   B/pos:    3e+02
Iter:     6000   Time(s):     9.44   MPos/s:     10.4   It/s:    6e+02   MB/s:    3e+03   B/pos:    3e+02

```

### DIFF
--- a/data_loader/cpp/training_data_loader_bench.cpp
+++ b/data_loader/cpp/training_data_loader_bench.cpp
@@ -27,8 +27,56 @@ g++ -std=c++20 -g3 -O3 -DNDEBUG -DBENCH -march=native \
 #include <fstream>
 #include <string>
 #include <thread>
+#include <filesystem>
 
 #include "training_data_loader_abi.h"
+
+namespace fs = std::filesystem;
+
+/**
+ * Validates existence, copies files to /dev/shm for low-latency access,
+ * and returns the new RAM-based paths.
+ */
+std::vector<std::string> stage_files_to_ram(int file_count, const char** files) {
+    std::vector<std::string> ram_paths;
+    ram_paths.reserve(file_count);
+
+    // Using /dev/shm for Linux RAM-disk.
+    // For cross-platform, you would need a custom memory buffer interface.
+    const fs::path ram_base = "/dev/shm/app_cache";
+
+    try {
+        if (!fs::exists(ram_base)) {
+            fs::create_directories(ram_base);
+        }
+
+        for (int i = 0; i < file_count; ++i) {
+            if (files[i] == nullptr) continue;
+
+            fs::path original_path(files[i]);
+
+            // 1. Logic/Existence Validation
+            if (!fs::exists(original_path) || !fs::is_regular_file(original_path)) {
+                throw std::runtime_error("Invalid or missing file: " + original_path.string());
+            }
+
+            // 2. Performance: Copy to RAM
+            fs::path target_path = ram_base / original_path.filename();
+
+            // Avoid redundant copies if multiple streams use the same files
+            if (!fs::exists(target_path)) {
+                fs::copy_file(original_path, target_path, fs::copy_options::overwrite_existing);
+            }
+
+            ram_paths.push_back(target_path.string());
+        }
+    } catch (const fs::filesystem_error& e) {
+        std::cerr << "Filesystem error: " << e.what() << std::endl;
+        throw;
+    }
+
+    return ram_paths;
+}
 
 long long get_rchar_self() {
     std::ifstream io_file("/proc/self/io");
@@ -58,6 +106,8 @@ struct SparseBatchStreamDeleter {
 int main(int argc, char** argv) {
     int concurrency = std::thread::hardware_concurrency();
     size_t iteration_count = 6000;
+    size_t warmup_iterations = 5;
+    int do_cache_files = 1;
 
     int i = 1;
     for (; i < argc; ++i) {
@@ -67,6 +117,8 @@ int main(int argc, char** argv) {
             concurrency = std::stoi(argv[++i]);
         } else if (arg == "-i" && i + 1 < argc) {
             iteration_count = std::stoul(argv[++i]);
+        } else if (arg == "-c" && i + 1 < argc) {
+            do_cache_files = std::stoi(argv[++i]);
         } else if (arg[0] == '-') {
             std::cerr << "Unknown option: " << arg << "\n";
             return 1;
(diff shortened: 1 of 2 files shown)


TonyCongqianWang:
> While it is straightforward to make changes there, it breaks backwards compatibility of recipes. So cleanups that do not break nettest I think are good, otherwise we have to weigh the pros and cons.

Got it. I can offer a compromise where I keep the `.bat` that simply calls the `.sh` and an extra step that copies the `.so` from  build to root.

Although I don't understand why nettest should care about the location of either... Doesn't it just use docker, which hides all of those implementation details?

vondele:
btw the CI check on training-data-loader is taking much too long..

data_loader/cpp/training_data_loader_bench.cpp
```diff
@@ -38,20 +43,56 @@ long long get_rchar_self() {
...
+};
 
+struct SparseBatchStreamDeleter {
+    void operator()(SparseBatchStream* s) const {
+        destroy_sparse_batch_stream(s);
+    }
+};
+
+int main(int argc, char** argv) {
 #ifdef PGO_BUILD
```
vondele: I guess we can get rid of this now with the options?

.github/workflows/build.yml
```diff
@@ -14,11 +14,16 @@ jobs:
...
         run: >
-          g++ -std=c++20 -g3 -O3 -DNDEBUG -DBENCH -DPGO_BUILD -march=native training_data_loader_bench.cpp training_data_loader.cpp
-
+          g++ -std=c++20 -g3 -O3 -DNDEBUG -march=native -fPIC -shared \
+              data_loader/cpp/training_data_loader.cpp \
+              -o build/libtraining_data_loader.so && \
+          g++ -std=c++20 -g3 -O3 -DNDEBUG -DBENCH -march=native \
+              data_loader/cpp/training_data_loader_bench.cpp \
+              -L./build -ltraining_data_loader -Wl,-rpath,'$ORIGIN/build' \
+              -o bench_shared
```
vondele: what about `data_loader_bench`

data_loader/cpp/training_data_loader_bench.cpp
TonyCongqianWang: I guess so, but I didn't want to break any default behaviour.. your call

.github/workflows/build.yml
TonyCongqianWang: What do you mean?

You mean adding instructions to use the cmake compiled program instead of the manually compiled one?

data_loader/cpp/training_data_loader_bench.cpp
vondele: well, it is basically an internal thing of the cmake build process, so the 'as-if' rule applies (assuming you adjust the cmake build process).

TonyCongqianWang: That's true, it should be an internal thing, although the old ci pipeline used the PGO flag directly

TonyCongqianWang:
Ok I resolved the error. Somehow there were some conflicting cmake settings. One of them is quite wild: PGO_USE actually set the build type to be deb with symbols.. which is definitely not what we want for performance. I suspect that it was used for debugging at some point

t0278611:
Last commit is optional and can be reverted..

TonyCongqianWang:
Can we get this merged?

PR #415, opened 2026-03-02
Improving DDP on HPC with Numa

TonyCongqianWang:
- fixing some small issues and inconsistencies 
- adding cpu affinity binding 
- limit core oversubscription 

To get best performance start with `torchrun --nproc_per_node=$NUM_GPUS` instead of directly with `python`

### DIFF
--- a/model/lightning_module.py
+++ b/model/lightning_module.py
@@ -38,7 +38,6 @@ class NNUE(L.LightningModule):
         num_psqt_buckets=8,
         num_ls_buckets=8,
         loss_params=LossParams(),
-        compile_backend=None,
     ):
         super().__init__()
         self.model: NNUEModel = NNUEModel(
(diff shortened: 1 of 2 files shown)


data_loader/cpp/training_data_loader.cpp
```diff
@@ -604,6 +604,7 @@ struct FeaturedBatchStream: Stream<SparseBatch> {
     ~FeaturedBatchStream() {
         m_stop_flag.store(true);
         m_batches_not_full.notify_all();
+        m_batches_any.notify_all();
```
vondele: if these changes to the dataloader are independent, as I think they are, please move them to a separate PR, and explain why they are needed?

TonyCongqianWang: Yes you are right. I do have separate PR though. Im gonna remove them from here

config.py
```diff
@@ -19,6 +19,12 @@ class TrainingConfig(LossParams, DataloaderSkipConfig, FeatureConfig, ModelConfi
     gpus: Optional[str] = None
     """List of gpus to use, e.g. 0,1,2,3 for 4 gpus. Default: None (Use device 0 only)."""
 
+    pin_memory: bool = True
+    """Whether to use pin memory in the data pipeline. Should generally be left on unless you encounter issues with too much RAM usage."""
+
+    data_loader_queue_size: int = 10
```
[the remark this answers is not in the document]
TonyCongqianWang: I Will just fix it to 16. Should be small enough to not cause OOM and it doesn't seem like larger queues absorb the fluctuations effectively -- which is slightly surprising

A queue of 512 should equate to something like 10 seconds of no data... It would mean that the queue either never really fills up (which would be weird given that small net trains much faster) or that the lags are longer than 10 seconds!

Is there any way you could actually measure that?

TonyCongqianWang:
Can we merge this? Seems to be working, right? Once we merged this we can safely iterate on the next bottleneck

vondele:
seems still to have conflicts with master, so would need a rebase?

maybe @xu-shawn can comment on the changes to the lightning module. I can't judge the need for those. 

The ddp related stuff seems to be fine.

TonyCongqianWang:
Ok I can rebase

TonyCongqianWang:
I have resolved the merge conflict and cleaned up the unwanted changes..

Just depending on whether we want the compile during setup or in the train script like before...

PR #422, opened 2026-03-07
Refactor Lighting Module Configuration and Optimizer Set

TonyCongqianWang:
Two major changes:

1. Refactoring configs across the project to be more consistent with tyro:
* Using nested configs for lightning module.
* Modernized `serialize.py` cli with tyro.

2. Refactoring lightning modules Optimizer setup. Added wrapper classes for optimizers so that swapping to different optimizers creates less friction.

* Adding schedulefree as possible optimizer.
* Default behavior is not changed

* Using ranger21 from pip instead of custom file. Note: after `pip install ranger21` using `diff /usr/local/lib/python3.12/dist-packages/ranger21/ranger21.py ranger21.py` reveals that reveals that the `ranger21.py` file in master is identical to the one installed by pip

This might be useful for anyone that wants to try schedulefree. Also it opens the path to add other optimizers in the future. Maybe the best overall recipe uses different optimizers between stages

The reason why I think it us useful to merge now even though schedulefree hasn't been successful so far is that it removes friction in the future (especially in terms of merge conflics) when we want to test other optimizers.

### DIFF
--- a/train.py
+++ b/train.py
@@ -192,10 +192,9 @@ def main():
             )
         # we can set the following here just like that because when resuming
         # from .pt the optimizer is only created after the training is started
-        nnue.loss_params = loss_params
         nnue.max_epoch = max_epoch
         nnue.num_batches_per_epoch = args.num_batches_per_epoch
-        nnue.nnue_lightning_config = args.nnue_lightning_config
+        nnue.config = args.nnue_lightning_config
         nnue.param_index = args.dataloader_config.param_index
 
     input_feature_name = nnue.model.input_feature_name


config.py
```diff
@@ -5,11 +5,12 @@
 
 from data_loader.config import DataloaderSkipConfig
 from model.config import LossParams, ModelConfig
+from model.optimizers.optimizer_config import OptimizerConfig
```
xu-shawn: could you export the classes through an `__init__.py` file under model/optimizers?

train.py
```diff
@@ -166,10 +166,14 @@ def main():
...
             feature_name=feature_name,
             loss_params=loss_params,
+            optimizer_name=args.optimizer_name,
             max_epoch=max_epoch,
             num_batches_per_epoch=num_batches_per_epoch,
             gamma=args.gamma,
+            ft_weight_decay=args.ft_weight_decay,
+            dense_weight_decay=args.dense_weight_decay,
             lr=args.lr,
+            warmup_steps=args.warmup_steps,
```
xu-shawn: if we were to add these parameters, it would be a good idea to group them into OptimizerConfig and do something like this to extract argparse results:
https://github.com/official-stockfish/nnue-pytorch/blob/e47dde7f9746c7522b58bb5c09016653888f4f05/data_loader/config.py#L27-L38

config.py
TonyCongqianWang: I could. But do we want them to be "private" or public?

train.py
TonyCongqianWang: I also thought about it... We could do composition instead of inheritance with the parameters... Would be weird if it's only here and inheritance everywhere else. I'd suggest it to be somewhat consistent

TonyCongqianWang: I am not sure If I understand you correctly though.. can you give me an example of what you think would be good?

 I am using a dataclass `OptimizerConfig` and tyro to extract the parameters

config.py
xu-shawn: imo it should make sense to export `OptimizerConfig`, `Ranger21Wrapper`, and `ScheduleFreeWrapper`

train.py
xu-shawn: I'm suggesting something like this:

```py
@dataclass
class OptimizerConfig:
    ...

    def get_optimizer_config_from_args(args) -> "OptimizerConfig"
         ...
         return OptimizerConfig(...)
```

And refactoring the `NNUE` class to take in an `OptimizerConfig` in the constructor instead of various separate optimizer parameters

Disservin: get_dataloader_skip_config_from_args is actually not really needed anymore.. you can replace the current call side with just args, since it inherits from DataloaderSkipConfig

xu-shawn: ah I guess then just refactoring the NNUE class constructor should be good

Disservin: yeah though it feels not so clean when you pass the entire object and rely on type hints, so im nto sure what to do

TonyCongqianWang: The best solution would be to use nested configs (composition) instead of inheritance. Tyro can handle them very well.

Then you can indeed pass the optimizer config class which would the best.

But I would suggest doing all of the config refactoring in one PR

config.py
TonyCongqianWang: Ok sure. Can do.

I'm gonna work on it tomorrow

train.py
Disservin: yea see https://github.com/official-stockfish/nnue-pytorch/pull/425

config.py
TonyCongqianWang: Done, please review :)

train.py
TonyCongqianWang: Done, please review. Probably gonna throw out argparse at serialize.py as well but I just wanted something worked before going to bed...

serialize.py
```diff
@@ -1,15 +1,19 @@
...
 import model as M
 
-
 def main():
-    parser = argparse.ArgumentParser(
-        description="Converts files between ckpt and nnue format."
+    class NoExitParser(argparse.ArgumentParser):
+        def error(self, message):
+            # Instead of exiting, we just record that it failed or ignore it
+            print(f"Argparse ignored: {message}")
```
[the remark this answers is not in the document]
TonyCongqianWang: I will let Gemini write me a tyro class to get rid of the argparse hack

train.py
```diff
@@ -203,14 +191,12 @@ def main():
...
-        nnue.num_batches_per_epoch = num_batches_per_epoch
         # we can set the following here just like that because when resuming
         # from .pt the optimizer is only created after the training is started
-        nnue.gamma = args.gamma
-        nnue.lr = args.lr
-        nnue.compile_backend = args.compile_backend
+        nnue.loss_params = loss_params
+        nnue.max_epoch = max_epoch
+        nnue.num_batches_per_epoch=args.num_batches_per_epoch
+        args.nnue_lightning_config = args.nnue_lightning_config
```
Disservin: args.nnue_lightning_config = args.nnue_lightning_config ?

serialize.py
```diff
@@ -95,65 +85,65 @@ def main():
     elif args.source.endswith(".nnue"):
         with open(args.source, "rb") as f:
             nnue = M.NNUE(
-                feature_name,
-                M.ModelConfig.get_model_config(args),
-                M.QuantizationConfig(),
+                config=nnue_lightning_config,
+                quantize_config=M.QuantizationConfig(),
             )
```
Disservin, line 90: what happened to feature_name?

train.py
TonyCongqianWang: Oops...

Will fix! Thanks for the review

serialize.py
TonyCongqianWang: I is now part of the model config

Disservin: nvm

train.py
```diff
@@ -173,26 +169,18 @@ def main():
...
 
-    feature_name = args.features
+    feature_name = args.nnue_lightning_config.features
 
     print("Loss parameters:")
     print(loss_params)
 
-    num_batches_per_epoch=max(
-                1, args.epoch_size // global_batch_size_requested
-            )
```
Disservin, line 183: is this correct that this is gone now?

Disservin: ah i see it's a property okay

TonyCongqianWang: Yes!

It is moved inside the training config as property as it can be calculated from batch size and epoch size, both part of training config

model/optimizers/ranger21_wrapper.py
```diff
@@ -0,0 +1,68 @@
...
+
+    num_batches_per_epoch: int = 100_000_000 // 16384
+    """Number of batches per epoch, used by Ranger21 for scheduling."""
+
+class Ranger21Wrapper:
+    def __init__(
+        self,
+        config,
+        max_epoch,
+        num_batches_per_epoch,
```
xu-shawn, line 23: is this argument ever used?

model/optimizers/ranger21_wrapper.py
```diff
@@ -0,0 +1,68 @@
...
+    ):
+        self.max_epoch = max_epoch
+        self.num_batches_per_epoch = config.num_batches_per_epoch
```
xu-shawn: if I understand correctly, `num_batches_per_epoch` is now calculated in `TrainingConfig` and passed via function/constructor parameters, but here, `config.num_batches_per_epoch` is the default value from `OptimizerConfig`, which is inherited from `Ranger21Config` and never modified

model/optimizers/ranger21_wrapper.py
```diff
@@ -0,0 +1,68 @@
...
+    _ranger21_import_error = False
+except ImportError:
+    _ranger21_import_error = True
+
+@dataclass
+class Ranger21Config:
+    gamma: float = 0.992
+    """Multiplicative factor applied to the learning rate after every epoch."""
+
+    num_batches_per_epoch: int = 100_000_000 // 16384
```
xu-shawn: should this be removed?

model/optimizers/ranger21_wrapper.py
TonyCongqianWang: Honestly I don't know.. we could look inside the ranger21 code.

As warmup and warm down are both disabled it might be unused.

Master branch does pass it regardless however

model/optimizers/ranger21_wrapper.py
TonyCongqianWang: Oops, this is definitely a mistake yes. Will fix

model/optimizers/ranger21_wrapper.py
xu-shawn: oh I meant that it's not used in the constructor. I think I elaborated on it more clearly in the review below, so I'll just resolve this

TonyCongqianWang: Ahh I see haha. Misunderstanding.

But I did look at the code.. and it does seem like num_batches_per_epoch is essentially unused by ranger21 in our setup...

config.py
```diff
@@ -82,10 +74,19 @@ class TrainingConfig(FeatureConfig):
...
-    model_config: OmitArgPrefixes[ModelConfig] = field(default_factory=ModelConfig)
+    nnue_lightning_config: OmitArgPrefixes[NNUELightningConfig] = field(
+        default_factory=NNUELightningConfig
+    )
 
-    loss_config: OmitArgPrefixes[LossParams] = field(default_factory=LossParams)
+    @property
+    def batch_size(self) -> int:
+        """Returns the validated batch size."""
+        return self.batch_size_raw if self.batch_size_raw > 0 else 16384
```
xu-shawn: is this check now unnecessary? batch_size_raw is now 16384 by default

train.py
```diff
@@ -203,14 +191,12 @@ def main():
...
-        nnue.max_epoch = max_epoch
-        nnue.num_batches_per_epoch = num_batches_per_epoch
         # we can set the following here just like that because when resuming
         # from .pt the optimizer is only created after the training is started
-        nnue.gamma = args.gamma
-        nnue.lr = args.lr
-        nnue.compile_backend = args.compile_backend
+        nnue.loss_params = loss_params
+        nnue.max_epoch = max_epoch
+        nnue.num_batches_per_epoch=args.num_batches_per_epoch
```
xu-shawn: minor formatting inconsistency. did you run ruff format?

config.py
TonyCongqianWang: That's true. Should we just throw an error if someone tries to pass something <=0?

train.py
TonyCongqianWang: Oh lol didn't know about it. Relied on vcscode so far.... Which doesn't handle python super reliably 

And also the GitHub automated workflow of course. So it seems ruff didn't flag thig inconsistency. Good spot

config.py
TonyCongqianWang: Turns out input validation with tyro is surprisingly difficult. There seems to be a bug or wrong documentation for the custom constructors.

Also surprisingly there is no equivalent to nargs='+' in tyro such that for positional arguments of type list, an empty list is always valid

PR #428, opened 2026-03-08
Data loader memory management

TonyCongqianWang:
I have no idea if this is faster on the cluster perhaps. Locally I seem to be bounded by 4GB/s but I am unsure where it comes from. I suspect that one mutex is the main bottleneck and essentially bounds it to single thread reading speed, which might be 4GB/s.

Maybe @vondele you can find out more through experiments...

I added jemalloc to the docker and using the wrapper script
`./scripts/run_with_custom_malloc.sh ./build/training_data_loader_bench .pgo/small.binpack` you can inject it to the benchmark. Which might be faster... or not. locally there was zero difference as everything hits the magical 4GB/s wall


TonyCongqianWang:
Hmm we could keep the libjemalloc and try nettest with it. Maybe it does speed things up overall even if the loader benchmark itself doesn't change

PR #430, opened 2026-03-08
refactor skip predicate logic

TonyCongqianWang:
Improves the skip predicates efficiency and configurability.

positions with 3-4 pieces represent a non-trivial fraction of the training data (> 4%) even though they are extremely non-informative.
One symptom is that late game positions make up a very large portion as well (>16% are 100+ plies).

With the current quadratic interpolation, it is impossible to remove this behavior without underrepresenting useful positions with 6-14 pieces. I increased the granularity of the piece count config.

For the early plies, a soft filter is applied on top of the hard filter, allowing for a more gradual change.

Additionally Copilot pointed out in an earlier PR that the alpha calculation was wrong.

Instead of looping through piece count and finding the most represented piece count, it just calculated the current piece count overrepresentation 33 times. If for some reason the current piece count at the point of the update was overrepresented, the skip rate would skyrocket unintentionally.

Here this behavior is changed to allow for a more smooth alpha. Crucially, the skipping rate should stay somewhat consistent instead of wildly fluctuating.

Additionally some performance optimizations are implemented.

It must be tested if this new skip config improves training results or not


data_loader/cpp/training_data_loader.cpp
```diff
@@ -622,98 +622,208 @@ FenBatch* FenBatchStream::next() {
...
+        static thread_local bool last_score = VALUE_NONE;
+
+        // skip when 0 was used as placeholder for unavailable value
+        // detected through heuristic: Game was not a draw
+        // and last value was somewhat large
+        bool skip_placeholder_zero =
+            e.ply > last_ply &&
+            last_score != VALUE_NONE &&
+            std::abs(last_score) > 100 &&
+            e.result != 0;
```
TonyCongqianWang: Wow I forgot to add `e.score == 0` .... considering that, it did surprisingly well I guess...

data_loader/cpp/training_data_loader.cpp
```diff
@@ -622,98 +622,208 @@ FenBatch* FenBatchStream::next() {
...
+            early_ply_accept_prob[i] = interpolate_ply(static_cast<double>(i));
+        }
+    }
+
+    return [config, random_skip_threshold, target_pc_weights_lut, target_pc_weights_total,
+            early_ply_accept_prob = std::move(early_ply_accept_prob)](const TrainingDataEntry& e) {
+
+        static constexpr int VALUE_NONE = 32002;
+        static thread_local int last_ply = -1;
+        static thread_local bool last_score = VALUE_NONE;
```
TonyCongqianWang: wow this mistake is even more ridiculous. I accidentally used bool here instead of int... So actually it just never skipped!

TonyCongqianWang:
https://tests.stockfishchess.org/tests/live_elo/69f046ba1e5788938e86a42f

Passed fishtest!

xu-shawn:
Congratulations! Could you add some lite documentation on what the new control points do?

data_loader/cpp/training_data_loader.cpp
```diff
@@ -622,98 +622,217 @@ FenBatch* FenBatchStream::next() {
...
+            accept_prob = alpha * current_ratio;
+        }
 
-            return false;
-        };
-    }
-    return nullptr;
+        accept_prob = std::clamp(accept_prob, 0.0, 1.0);
+
+        uint64_t reject_threshold = static_cast<uint64_t>((1.0 - accept_prob) * static_cast<double>(~0ULL));
```
xu-shawn: Using `std::bernoulli_distribution` here should be identical?

TonyCongqianWang: It is equivalent logically but the implementation does make quite a difference in performance. See  https://github.com/llvm/llvm-project/issues/144237. The reason for this is that Bernoulli does quite some sanity checks and also has higher resolution, which really is a bit overkill for the data loader

TonyCongqianWang:
> Congratulations! Could you add some lite documentation on what the new control points do?

Thank you!

I elaborated more on the doc strings. Could you take a look? Is the description understandable for you?

Disservin:
i pushed a clang format commit, i think to me this looks ready ? can merge this before sf merges the net to just reduce the merge queue

vondele:
LGTM

PR #431, opened 2026-03-09
Reducing locks in data_loader / parallel_file_reader

TonyCongqianWang:
This PR removes three major bottlenecks that harm data_loader performance on HPC systems:
- global lock on file read: Only one thread was able to read from filesystem at the same time which is fine on local SSDs but catastrophic for high latency network drives.
- global lock on buffer read/write: All workers shared a single buffer and all readers had to fight to swap their buffer with the shared buffer. The buffer could only be replaced once empty, until then all reader threads had to wait once there buffer was full. All worker threads were grabbing from the same buffer and technically they never acquire a lock (fast) but it was also a race condition (bad).
- worker threads didn't actually do much work as profiling revealed: reader threads were doing most of the work (mostly decompressing) and worker threads were idling.

The Bottlenecks were solved as such:
- striped locking: instead of one global mutex, one mutex is used per file
- ring buffer queue: instead of having one single buffer there is a small ring buffer that lets readers swap their buffer and continue reading quickly. All worker threads receive their own buffer as well, limiting access to shared resources while being thread safe.
- A simple change of reader / worker ratio significantly improved performance.

### DIFF
--- a/data_loader/cpp/lib/nnue_training_data_formats.h
+++ b/data_loader/cpp/lib/nnue_training_data_formats.h
@@ -7791,6 +7791,7 @@ namespace binpack
                     return this->should_stop_consumer();
                 });
                 if (!success) return std::nullopt;
+                local.offset = 0;
             }
 
             return std::move(local.entries[local.offset++]);
--- a/data_loader/cpp/lib/nnue_training_data_stream.h
+++ b/data_loader/cpp/lib/nnue_training_data_stream.h
@@ -9,6 +9,8 @@
 #include <memory>
 #include <vector>
 #include <functional>
+#include <atomic>
+#include <mutex>
 
 namespace training_data {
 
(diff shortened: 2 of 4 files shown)


Sopel97:
Good changes regarding the locks, we missed this opportunity when adding the ability to interleave reads from multiple files.

It's unclear to me if the gains are mainly from finer-grained locking or from the buffer allocation changes. This should be split into two PRs.

One thing that bothers me is those leaked-by-design allocations. I don't see where the destruction order would be problematic here?

Sopel97:
Locking per file should be like 5-10 lines of code.

Then we can evaluate if whatever the ring buffers gain are worth the additional complexity and leaks.

btw. what's the race in the current code?

Sopel97:
the swap is done under a lock, and the producer is not touching the m_buffer

PR #439, opened 2026-03-18
Freeze params

TonyCongqianWang:
Splits weights from feature transformer into psqt and ft weights. This adds a lot of flexibility at a slight speed penalty.

However that flexibility can be leveraged to freeze psqt weights which gains speed.


TonyCongqianWang:
Probably not useful anymore.. maybe in the future

PR #442, opened 2026-04-01
Quant fix

TonyCongqianWang:
Improves quant config and removes hard coded values from the code.

The main goal is to improve readability and maintainability. As of currently there is a mismatch between training and inference (see #416 for more information). This patch intends to not only fix the mismatch but also to prevent mismatches in the future by inferring runtime correction factors directly from the quantization config instead of relying on manual changes across the codebase.

This allows arbitrary changes to the weight and activation quantization scales without having to make changes across the entire codebase.

As a quality of life improvement, serialize.py and quantize.py were refactored and the information density of the histogram callback was improved: The current histogram callback reports three values that are rather uninformative, since the model weights are already clipped through the callbacks. The logic was changed to use the native clipping of the model first and then clip only for type safety for serialization.

Changes quant scheme slightly (Test code at https://github.com/TonyCongqianWang/Stockfish/tree/gruenerapfel/quant_fix).

In particular ft_one = 256 and  hidden_one=128 for clean power of two scales -- which is the standard for fixed scale qat.

Output scale is handled by https://github.com/TonyCongqianWang/Stockfish/blob/384175653285f68b67bd6a8d696380c40731e8c8/src/nnue/nnue_architecture.h#L135 which should yield slightly higher resolution. The main advantage however is better maintainability.

The maximum number of fully saturated inputs to i32 with i8*i8 inputs would be `133,144 / (600 * OutputScale) = 13` There is actually some risk of overflow even with the current setup, as even L3 has enough neurons and L1 has a skip connection. However given that L1=8k worked, this should be fine. But as a safety measure I will change the order of operation so that the maximum number of fully saturated neurons becomes `133,144 / (600 * OutputScale / WeightScale) = 887`

### DIFF
--- a/model/quantize.py
+++ b/model/quantize.py
@@ -28,7 +28,7 @@ def _safe_convert(value: torch.Tensor, target_dtype: torch.dtype):
 @dataclass
 class QuantizationConfig:
     nnue2score: float = 600.0
-    weight_scale_hidden_0: float = 64.0
+    weight_scale_hidden_0: float = 128.0
     weight_scale_hidden_1: float = 64.0
     weight_scale_hidden_2: float = 128.0
     weight_scale_out: float = 16.0
--- a/model/quantize.py
+++ b/model/quantize.py
@@ -13,6 +13,17 @@ class WeightClippingConfig(TypedDict):
     max_weight: float
     virtual_params: NotRequired[torch.Tensor]
 
+def _safe_convert(value: torch.Tensor, target_dtype: torch.dtype):
+    _info = torch.iinfo(target_dtype)
+    # Symmetric range: [-max, max]
+    min_val = -_info.max
+    max_val = _info.max
+
+    rounded_value = value.round()
+    clamped_value = rounded_value.clamp(min_val, max_val)
+    num_clipped = (rounded_value != clamped_value).sum()
+    quantized_value = clamped_value.to(target_dtype)
+    return quantized_value, num_clipped
 
 @dataclass
 class QuantizationConfig:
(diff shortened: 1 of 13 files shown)


model/quantize.py
```diff
@@ -57,11 +78,18 @@ def generate_weight_clipping_config(
             },
             {
                 "params": [model.layer_stacks.output.linear.weight],
-                "min_weight": -self.max_out_weight,
-                "max_weight": self.max_out_weight,
+                "min_weight": -self.max_hidden_weight,
+                "max_weight": self.max_hidden_weight,
```
Sopel97: the last layer has different quantization, is this really correct here?

Sopel97: master is 1.68, this is 2

TonyCongqianWang: You were too fast 😭 I was about to edit the description

vondele:
> at the cost of slightly higher overflow risk.

I'm pretty sure we want to demonstrate it can't overflow, low 'risk' is probably not enough.

TonyCongqianWang:
I don't think we can demonstrate that the current net can't overflow can we?
It is save unless the out float value goes over like 27.

The largest contributor is the skip connection which can be as large as `L1 * max_weight_hidden`. The added amount now is bounded by ``L3 * max_weight_hidden`` lastly, the intermediate reduction was only ~0.85.

The risk of overflow is virtually zero, but if you know a way to guarantee safety, do elaborate.

For safety the sum could be calculated in i64 before doing the out scaling... Since it is only one value, the cost should be rather negligible.

Ah wait yes. It is guaranteed... The max weight and max activations are always `127` in the quantized version.

`127*127*(1024 + 32)` is easily below the i32 threshold. However we have to be careful if we decide to increase L1 *and* L3 by a lot in the future. But yeah.... If L1=8k is fine, then this is also fine

Sopel97:
but what is this trying to solve? the quantization scheme was arbitrary but mathematically sound. Now this just makes it more arbitrary (and possibly unsound with how max differs from one)?

TonyCongqianWang:
How so?
This is mathematically sound also *not* arbitrary as taking powers of two is standard practice

TonyCongqianWang:
reopened at #448

PR #444, opened 2026-04-02
Pawn pusher threats

anematode:
Companion to https://github.com/official-stockfish/Stockfish/pull/6694


xu-shawn:
Congrats on the gainer! I wonder if we should bump feature set version to FullThreatsv2?

PR #445, opened 2026-04-06
patch to prevent all threads to get stuck on the same slow file

TonyCongqianWang:
The striped lock for files proved to be a great efficiency boost for training on a cluster, although it could slightly skew the data distribution towards faster files.

Still if one filesystem hangs, all threads could eventually get stuck on the same mutex. This patch prevents this and also emits a warning so that there is no silent error


TonyCongqianWang:
@vondele Could you somehow test this on the cluster?

TonyCongqianWang:
I resolved all comments. If you could test it while the fileserver is slow, that would be appreciated

TonyCongqianWang:
Can you test this @vondele?

vondele:
I've been running this for a while now, this seems not slower than master and will protect in the exceptional case, so looks good to me.

PR #446, opened 2026-04-07
Smart bucket

TonyCongqianWang:
Allows net to learn dynamic routing to buckets.

Currently bucket 0 is for piececount 1-4 which is essentially useless


TonyCongqianWang:
Didn't seem to have learned good routing like this. Maybe some other approach is necessary

PR #447, opened 2026-04-08
add flag to skip affinity

Disservin:
not everyone benefits from multi gpu, so sometimes it's beneficial to use both gpus for different training runs, in which case affinity needs to be disabled


TonyCongqianWang:
This does surely work, but I don't really see the point. Disabling the numa binding basically completely disables the script besides ``os.environ["PYTHONUNBUFFERED"] = "1"`` and the core count enforement. Both things can be done manually too.

The benefit I see though is that this can keep a cleaner interface in case we add more ddp improvements later on.

Also note that usually you still want affinity even when only using one GPU per run.

related PR that tackles the issue from the nettest side (though only for one GPU) https://github.com/vondele/nettest/pull/282.

Still think its a solid change that improves future compatibility

PR #451, opened 2026-04-16
Experimental/sparse kernel improvements

TonyCongqianWang:


vondele:
so finally got to testing this PR. 

First, I think it should be split, the smart matrix precision and the other changes. In that case we can measure independently, and things will be easier to roll back if there are issues. 

First the unit tests. I ran them multiple times but they are pretty reproducible.

```
$ for i in `seq 1 3`; do python tests/test_fused_double_ft.py; done
>>> Running Mathematical Consistency Test...
Check           | Max Difference       | Status
--------------------------------------------------
Forward L1      | 2.98e-08             | PASSED
Forward WPSQT   | 0.00e+00             | PASSED
Forward BPSQT   | 0.00e+00             | PASSED
Weight Grad     | 4.77e-07             | PASSED
Bias Grad       | 2.38e-07             | PASSED

Consistency verified: Fused and Unfused modes are mathematically identical.

>>> Running Performance Benchmark for Training Step...
UNFUSED_EAGER     : 4,546,234 pos/s
UNFUSED_COMPILE   : 5,595,022 pos/s
FUSED_EAGER       : 6,898,836 pos/s
FUSED_COMPILE     : 6,964,825 pos/s

--- Summary ---
Compiler efficiency (Unfused): 1.23x
Hand-Fused vs Compiled-Unfused: 1.23x
```

```
$ for i in `seq 1 3`; do python tests/test_feature_transformer.py; done
Tests passed.
4354747.259106702 pos/s
Tests passed.
372011.25454131444 pos/s
```

Next I ran a comparison with master. 
```
==== checkout master =====
==== current sha: 2e3ffcbb4d12c0bfec9e3b70c1da864cd2354328 ====
==== compile data loader ====
==== train ====
master 1 67.66it/s
master 2 68.31it/s
master 3 68.23it/s
```

branch, without changes, and passing the additional option --use-fused-double-ft
```
==== checkout d998fe70b5430581c3a9cd25e39d5ca119e755f8 =====
==== current sha: d998fe70b5430581c3a9cd25e39d5ca119e755f8 ====
==== compile data loader ====
==== train with extra opts: --use-fused-double-ft ====
d998fe70b5430581c3a9cd25e39d5ca119e755f8 1 69.94it/s
d998fe70b5430581c3a9cd25e39d5ca119e755f8 2 69.93it/s
d998fe70b5430581c3a9cd25e39d5ca119e755f8 3 69.75it/s
```

branch, with smart matrix coded to False
```
# disable the smart matrix precisions.
==== checkout noSmart =====
==== current sha: 30860e6bb99878322b0fae11efa96719fe57ce67 ====
==== compile data loader ====
==== train with extra opts: --use-fused-double-ft ====
noSmart 1 69.17it/s
noSmart 2 68.90it/s
noSmart 3 69.22it/s
```

So, I think there is real measurable speedup. Generally, I wouldn't hide this behind an option. If we have good enough evidence that is a gain, I would keep the program flow simple, i.e. just use it.

For the matrix multiplication, I think we would need to do a full train to be confident that it is a gain.

TonyCongqianWang:
Nice, thanks! I will clean stuff up. I think leaving the normal sparse kernel is good for testing purposes (e.g. the unittest) also it seems that the speedup isn't consistent across different devices, having the option to disable makes sense

vondele:
image

That's a screenshot of the nsys result using ddp=4, basically zooming in on a single batch.

vondele:
with some nvtx markers pointing out the torch compiled regions etc.
image

TonyCongqianWang:
Looks interesting. Anything I should look at in particular?

vondele:
not that I can tell... I think interesting is that the fused_nnue_forward/backward do not account for so much of the time in this configuration, there is lots of 'small stuff' trailing.

TonyCongqianWang:
My swa branch at least shows that the optimizer does account for some of the bottleneck so maybe the improvement will be bigger with both together.

I'll clean this up first for merging and then we will see...

TonyCongqianWang:
superseeded by #494

PR #453, opened 2026-04-17
Decouple sparse linear function from CUDA

linrock:
Enables tasks that should not depend on CUDA. For example:
- cross checking evals
- converting checkpoints
- feature transformer tests

Add a pure-PyTorch implementation so the sparse feature transformer runs on CPU and MPS. Make the CuPy kernel import optional.

SparseLinearFunction.apply stays the public entry point and uses the CUDA kernel when available.

### DIFF
--- a/config.py
+++ b/config.py
@@ -50,6 +50,9 @@ class TrainingConfig:
     threads: int = -1
     """Number of torch threads to use. Default automatic (cores)."""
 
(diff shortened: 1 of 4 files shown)


TonyCongqianWang:
Some great changes. Lgtm. Requiring cuda for even basic development is quite annoying right now.

Maybe someone use a copilot review just in case

PR #456, opened 2026-04-18
copy datasets to gpu while fetching 

Disservin:
- small speedup of 2-3 it/s locally, matters a bit more on slow memory systems, increases memory requirements on the GPU a bit

on a box with 2 training runs happening at the same this, this makes one go at 17 it/s while the other one is at 11 it/s

https://github.com/vondele/nettest/pull/302

### DIFF
--- a/data_loader/dataset.py
+++ b/data_loader/dataset.py
@@ -1,5 +1,6 @@
 import threading
 import queue
+from dataclasses import dataclass
 
 import torch
 from torch.utils.data import Dataset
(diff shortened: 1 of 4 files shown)


TonyCongqianWang:
Great change, although it makes training on a 5070 with the correct batch size impossible

Disservin:
>    Great change, although it makes training on a 5070 with the correct batch size impossible

a) why ? that card has 12GB which is more than enough
b) you could still disable the memory pin (not sure how much of a difference that even made)

vondele:
>  b) you could still disable the memory pin (not sure how much of a difference that even made)

isn't memory pinning needed to have async transfer?

Disservin:
> > b) you could still disable the memory pin (not sure how much of a difference that even made)
> 
> isn't memory pinning needed to have async transfer?

yes but I meant that you can still disable it to reduce vram usage

vondele:
ok, I think the benefit of the speedup offsets the increased memory needs.

PR #462, opened 2026-04-25
Enable training with CPU and Metal

linrock:
Gets training working at all, though not optimized yet.
- generalize some previously cuda-specific areas
- fix metal training, dense fallback for sparse linear, no torch.compile


vondele:
I also checked this doesn't influence the speed in the GPU case, and did have a CPU based training of our master net (only 3 epochs). So looks good to me.

PR #463, opened 2026-04-26
Improving containerization and CI/CD

TonyCongqianWang:
Enable full training pipeline incuding serialization on CPU and adds CI tests:
- Added dockerfile for CPU
- Added setup instructions for Apple Silicon in Readme
- Removed cuda and cupy in code paths when using `--no-cupy` or `--{acceleator|device}={cpu|mps}` from `serialize.py` and `ftperm.py`
- Added a test that runs a short training pipeline on either cuda, cpu or mps
- Added GitHub workflows to automatically run a short training pipeline and cross check eval on cpu

Fixes some issues and improves usability with `run_docker.sh`
- allows completely non-interactive execution
- fixes a bug that caused the container to misbehave if run by users with uid not in {0, 1000} (`docker run --user ...` in run_docker but only uid and gid 0 and 1000 exist in the container).


TonyCongqianWang:
Would be nice if ft_perm also supports mps.

Is there any particular reason we use cupy there instead of torch?

vondele:
> Is there any particular reason we use cupy there instead of torch?

I think just legacy, if it can be done easily in torch, why not.

TonyCongqianWang:
@linrock could you perhaps test on apple?

And see if I can add anything to `setup_mps.h`? I am not sure what the exact requirements are to use this

TonyCongqianWang:
CI seems to be running now 🥳

ftperm.py
```diff
@@ -184,108 +200,68 @@ def apply_rotate_right(perm: npt.NDArray, indices: tuple[int, ...]) -> None:
...
+        (shape[0], shape[1] // ZERO_BLOCK_SIZE, ZERO_BLOCK_SIZE)
     )
 
-    if use_cupy:
-        # Calculate number of zeros in each block
-        num_zeros = cp.sum(actmat_chunked, axis=2, keepdims=True)
-        # Broadcast back to the same shape as actmat_chunked so it's easier to work with
-        num_zeros = cp.tile(num_zeros, (1, 1, ZERO_BLOCK_SIZE))
-
-        # Marks an element if all other elements in a block are zero.
```
xu-shawn, line 205: maybe the comment block here shouldn't get deleted?

ftperm.py
```diff
@@ -184,108 +200,68 @@ def apply_rotate_right(perm: npt.NDArray, indices: tuple[int, ...]) -> None:
...
     for actmat_batch in batched(actmat, BATCH_SIZE):
-        swapped_zero_count += get_swapped_zero_positive_count(
-            actmat_batch, use_cupy=use_cupy
-        )
+        swapped_zero_count += get_swapped_zero_positive_count(actmat_batch)
 
     # (L1/2) x (L1/2)
-    if use_cupy:
-        # Subtract from each i-th slice the positive value of the current i-th placement.
-        # This is the place where we account for how much "bad" it would do.
```
xu-shawn, line 261: here as well

ftperm.py
```diff
@@ -184,108 +200,68 @@ def apply_rotate_right(perm: npt.NDArray, indices: tuple[int, ...]) -> None:
...
-        #
-        rest_zero_indicator = (
-            (num_zeros - actmat_chunked == ZERO_BLOCK_SIZE - 1)
-            .reshape(shape)
-            .astype(cp.int8)
-        )
-
-        # Sum all possible pairs of elements in a single sample of actmat_flat and rest_zero_indicator.
-        # Aggregate sum over the whole batch.
-        # This tells us how much "good" a swap of i-th and j-th slices would do. It doesn't consider
```
xu-shawn, line 222: here

TonyCongqianWang: readded 👍

ftperm.py
TonyCongqianWang: oops my bad. Wasn't intended. i added it again

linrock:
> @linrock could you perhaps test on apple?

i tried the training command from `cpu_testrun.yml` and it ran successfully

TonyCongqianWang:
Cross eval now also works!

TonyCongqianWang:
@Disservin @linrock @xu-shawn I'd say this is ready now. I improved the CI a bit so that it doesn't run when PRs are still in draft mode and also it ci can be skipped if the commit message contains `[wip]`.

Are we waiting for @vondele ? He is afk for the week though

TonyCongqianWang:
@vondele how shall we proceed. Can you request another copilot review?

I could split this in 2 + 1 PRs or clean the commits.

But it's not possible to split it into disjoint PRs I think

vondele:
Indeed, the PR is fairly large and if it can be split that would be good, I'll leave the final judgement on how to do it best to @xu-shawn. From looking at it, would it make sense to split in two pieces:
* whatever is needed to get the training, conversion, cross-check to run on cpu/mps
* a follow up PR that uses that to enable CI

vondele:
Given https://semgrep.dev/blog/2026/malicious-dependency-in-pytorch-lightning-used-for-ai-training/ .... should we specify explicit versions/shas in the container files / pip install ? This is once again separate from this PR.

TonyCongqianWang:
> Given https://semgrep.dev/blog/2026/malicious-dependency-in-pytorch-lightning-used-for-ai-training/ .... should we specify explicit versions/shas in the container files / pip install ? This is once again separate from this PR.

Probably a good idea for another PR yes...

I will fix the comments here first and then wait for @xu-shawn's decision and then split up the PR accordingly. It can be reasonably split up into two independent PRs:

1. adding the testing script that runs a short test pipeline for any device + all changes it depends upon
2. improving run_docker.sh as well as adding the CPU docker and mps README
3. workflow changes

xu-shawn:
Agree with vondele. One PR for cpu/mps compatiblility and one for CI looks the most reasonable. I also see some trivial cleanups like at the top of ftperm.py. If you'd like a third PR with these can both reduce the clutter and get merged right away.

TonyCongqianWang:
splitting this up into 2 + 1 PRs:
#464
#465 
#466

PR #466, opened 2026-05-02
Improve GitHub workflow

TonyCongqianWang:
- overhaul workflows by adding an orchestrator file that coordinates ci jobs
- added ci job that runs a short training pipeline and cross_check_eval
- prevents ci jobs other than ruff to be run on draft PRs or commits marked by [wip]
- added capability to build engine from github `owner/sha` in `cross_check_eval.py` for workflow compatibility

needs #464 and #465


docker_entrypoint.sh
```diff
@@ -0,0 +1,73 @@
...
+            # nnue_user exists but has the wrong UID
+            usermod -u "$TARGET_UID" -g "$TARGET_GID" -d "$HOME" -s /bin/bash nnue_user
+        else
+            # Neither the UID nor the name exists
+            useradd -u "$TARGET_UID" -g nnue_group -d "$HOME" -s /bin/bash nnue_user
+        fi
+    fi
+
+    # Fix ownership of the mount point if it was created by a different root process
+    chown "$TARGET_UID:$TARGET_GID" "$HOME"
```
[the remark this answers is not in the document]
TonyCongqianWang: already fixed in #464

run_docker.sh
```diff
@@ -4,51 +4,162 @@ set -e
...
+  echo "Enter the path to your data directory to mount into the container: "
+  read -r DATA_PATH
+fi
 
-DATA_PATH=${DATA_PATH}
 echo "Using data path: $DATA_PATH"
 
-# Checking if docker is in rootless mode.
+# 5. Handle user mapping and container home directory setup
+CONTAINER_HOME="$(pwd)/.container_home"
```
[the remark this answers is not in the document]
TonyCongqianWang: Id rather use .dockerignore here

.github/workflows/cpu-testrun.yml
```diff
@@ -0,0 +1,20 @@
...
+
+      - name: Perform Test Run of entire Training Pipeline (CPU)
+        run: |
+          ./run_docker.sh CPU ${{ github.workspace }} --non-interactive --exec \
+            python -u tests/test_training_pipeline_run.py -y --device cpu --test-dir /data/logs/training/runs/unittests_train_pipeline_cpu
+
+      - name: Perform Cross-Check Evaluation (CPU)
+        run: |
+          ./run_docker.sh CPU ${{ github.workspace }} --skip-setup --non-interactive --exec \
+            python -u cross_check_eval.py --net /data/logs/training/runs/unittests_train_pipeline_cpu/lightning_logs/version_2/checkpoints/last.nnue --data .pgo/small.binpack --engine /data/tmp/stockfish --build_engine_from_sha='' --device='cpu'
```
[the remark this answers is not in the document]
TonyCongqianWang: again Id rather add .dockerignore and keep the tests seperatre

ftperm.py
```diff
@@ -634,36 +631,44 @@ def ft_permute(model: NNUEModel, ft_perm_path: str) -> None:
...
-    BATCH_SIZE = 1000
+    BATCH_SIZE = 1024
 
     quantized_model = copy.deepcopy(model)
     quantize_ft(quantized_model)
-    quantized_model.cuda()
-
-    fen_batch_provider = make_fen_batch_provider(dataset, BATCH_SIZE)
+    if device_str != "cpu":
+        quantized_model.to(device_str)
```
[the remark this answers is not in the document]
TonyCongqianWang: will fix in #465

ftperm.py
```diff
@@ -674,7 +679,7 @@ def command_gather(args: FeaturePermutationConfig) -> None:
     if args.subcommand.checkpoint:
         nnue = NNUE.load_from_checkpoint(
             args.subcommand.checkpoint,
-            feature_name=args.feature_config.features,
+            feature_name=args.subcommand.feature_config.features,
             config=args.model_config,
             quantize_config=QuantizationConfig(),
```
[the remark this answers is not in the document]
TonyCongqianWang: will fix in #465

tests/test_training_pipeline_run.py
```diff
@@ -0,0 +1,91 @@
...
+            if response == 'y':
+                print(f"Deleting {test_dir}...")
+                shutil.rmtree(test_dir)
+            else:
+                print("Aborting. Directory must be removed before running to preserve version increment logic.")
+                sys.exit(1)
+
+    # --- 4. Define Pipeline Commands ---
+    pipeline = [
+        f"python -u train.py ./.pgo/small.binpack --batch-size 1024 --l1=1024 --features=Full_Threats+HalfKAv2_hm^ --epoch-size 10000 --max_epochs=2 --default_root_dir {test_dir_str} {train_device_arg} {train_workers_arg}",
```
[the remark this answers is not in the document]
TonyCongqianWang: will fix in #465

TonyCongqianWang:
Fixed comments, but better to first merge #464 and #465 before reviewing again

cross_check_eval.py
```diff
@@ -38,13 +47,86 @@ class CrossCheckConfig:
...
+
 
 @dataclass(frozen=True)
 class CliConfig:
     cross_check_config: OmitArgPrefixes[CrossCheckConfig]
     nnue_lightning_config: OmitArgPrefixes[M.NNUELightningConfig]
 
 
+
+def clone_and_build_engine(engine_dest_path: str, repo_info: str, force_build: bool):
```
vondele: I don't think this script should be building engines.

vondele:
please rebase on master after the latest merges

cross_check_eval.py
TonyCongqianWang: Hmm yeah you are right. This is a bit too hacky. It can easily be put in a standalone script... though where should I put it? inside .github? after all its not needed by anything else..

vondele: maybe just as a step in the cpu-testrun.yml?

vondele: Like with a checkout action, get SF, like here an example with matetrack:

```
      - name: Checkout matetrack repo
        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
        with:
          repository: vondele/matetrack
          path: matetrack
          ref: 6c8405fac9028ca66a077f5c96c918fec0ef8d1d
          persist-credentials: false
```

after that just build there

TonyCongqianWang: Ahh .... okay thats much easier...

.github/workflows/ci-orchestrator.yml
```diff
@@ -0,0 +1,89 @@
...
+  # Evaluate conditions for other jobs
+  evaluate-execution:
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master' || github.base_ref == 'master'
+    outputs:
+      run_secondaries: ${{ steps.check_status.outputs.run_status }}
+      run_data_loader_test_paths: ${{ steps.filter.outputs.data_loader_paths }}
+      force_all: ${{ steps.check_status.outputs.force_all }}
+    steps:
+      - uses: actions/checkout@v4
```
vondele: use explicit sha for actions

.github/workflows/ci-orchestrator.yml
```diff
@@ -0,0 +1,89 @@
...
+        with:
+          fetch-depth: 2
+
+      - name: Evaluate Path Restrictions
+        uses: dorny/paths-filter@v3
```
vondele: idem

.github/workflows/cpu-testrun.yml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  build-and-bench:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Checkout Compatible Stockfish Repository
+        uses: actions/checkout@v4
```
vondele: idem

.github/workflows/cpu-testrun.yml
```diff
@@ -0,0 +1,33 @@
+name: cpu-testrun
+
+on: [workflow_call]
+
+jobs:
+  build-and-bench:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
vondele: idem

.github/workflows/ci-orchestrator.yml
vondele: also use v6.0.2 : de0fac2e4500dabe0009e67214ff5f5447ce83dd

.github/workflows/ci-orchestrator.yml
```diff
@@ -0,0 +1,89 @@
...
+      )
+    uses: ./.github/workflows/data-loader-test.yml
+
+  call-cpu-testrun:
+    needs: [call-ruff-pr, evaluate-execution]
+    # Runs if master/draft conditions are met
+    if: >
+      needs.evaluate-execution.outputs.force_all == 'true' ||
+      needs.evaluate-execution.outputs.run_secondaries == 'true'
+    uses: ./.github/workflows/cpu-testrun.yml
```
Disservin: please lets just get rid of this.. there's no need for an "orchestrator", just add the triggers back to the other ci files, and add paths etc
```
on:
  pull_request:
    types: [opened, synchronize, reopened, ready_for_review]
    branches: [master]
    paths:
      - "data_loader/**"
      - "compile_data_loader.sh"
```

Disservin: and this custom commit message parsing for force ci and wip we also don't need

Disservin (changes requested):

.github/workflows/ci-orchestrator.yml
TonyCongqianWang: I have to disagree, but am open to hear your reasoning.

It is common for workflows to use orchestration.

The cpu pipeline takes over 5 minutes. CI runs are "free" but still have costs.

It is completely unnecessary to run it on every single push

Disservin: > It is common for workflows to use orchestration.

maybe when it's required but this here adds so much logic which is not at all required

> The cpu pipeline takes over 5 minutes. CI runs are "free" but still have costs.

if you want to avoid too many ci runs happening in parallel use, https://stackoverflow.com/a/72408109/12614623 otherwise the ci should always run to show the state of the pr to the maintainer

---

This pr like many others before adds way too much code again for what it is doing, which seems to be a recent trend when using LLM to write patches.

You also added features which weren't even documented in the PR description like "force ci".. no one would use that since it's not documented in the PR description.. however people will generally not even use "force ci" or "wip" when they make pr's because they won't go through to the project history or ci files to figure this out..

Github support skipping CI natively anyway, so adding "wip" is not at all necessary..
https://docs.github.com/en/actions/how-tos/manage-workflow-runs/skip-workflow-runs

TonyCongqianWang: Two things that I think are quite useful but not addressed:

1. No CI on drafts -- I don't see the reason to run CI on every single push in drafts (one could put `[skip ci]` in every commit sure. But it does seem quite unpractical).

2. No CI when ruff fails. Kind of wasteful to run another CI just because some lint failed. Skipping on the correction doesn't make sense and cancelling the running one is an incomplete workaround. As it still wastes resources.

3. As you mentioned, the workflow should always run before merging. This is not achievable when only using the path field. As it would always skip the workflow without any way to force it


I agree that it's a bit much code and if you have a better way of fixing the issues, I'd be happy to hear suggestions.

Otherwise I can also split the PR again. But I don't agree that it's "too much for what it does".

Disservin: > This is not achievable when only using the path field. As it would always skip the workflow without any way to force it

if the patch doesn't touch files covered by the path field, then we don't need a workflow run?

> No CI when ruff fails. Kind of wasteful to run another CI just because some lint failed. Skipping on the correction doesn't make sense and cancelling the running one is an incomplete workaround. As it still wastes resources.

I don't agree with this, you should see all possible issue's immediately.. because if you just make the patch and come back later all you see is the ruff failure and you fix it thinking everything works, but then you get failures from the others ci and you have to make a second commit fixing those.. it's more DX friendly to just run all the CI's and let the developer fix all issues at once

> No CI on drafts -- I don't see the reason to run CI on every single push in drafts (one could put [skip ci] in every commit sure. But it does seem quite unpractical).

it's more of a rare thing that people open drafts and constantly push changes there and if you really insist on we can use `if: github.event.pull_request.draft == false` on the job directly.. 
```
jobs:
  ci:
    if: github.event.pull_request.draft == false
```

TonyCongqianWang: Alright. I think adding the draft exemption and parallel run limitation is fine by me.

What formatting tool are you using?

I have been primarily using the ci ruff. Which is probably an anti-pattern. Maybe we can add some things about formatting tools/guidelines to the README?

vondele: just for the record, I agree with Disservin, simple PRs and simple code in the repo should be the goal, no additional features unless there is a clear need. The largest cost in software is in its maintenance, and complexity is a big barrier for new contributors to start. Getting that balance right is not easy though, so no worries, just take suggestions at heart.

Disservin: > What formatting tool are you using?

well ruff ? :D

TonyCongqianWang: I am a but confused since ruff -- at least the way it's used in the ci misses some of the formatting stuff. In particular line breaks

vondele: also my editor is set to use ruff.

Disservin: @TonyCongqianWang got an example ? if it's regarding windows linebreaks then your git isn't setup correctly

TonyCongqianWang: Hmm maybe I am just using it incorrectly.

I remember @vondele once touched a PR and fixed some `function(arg, arg ... arg)` formatting and it touched like 10 files ☠️

Disservin:
updated with what I had in mind.

TonyCongqianWang:
Yeah unlucky timing

PR #473, opened 2026-05-11
Adding new loss params: lambda jitter

TonyCongqianWang:
Adds perturbations to the lambda parameter used for loss calculation.

Perturbation amount is controlled by newly added parameters.

This allows recipes to work without a lambda schedule (i.e. with lambda_start = lambda_end).

Despite the name, this is the "secret sauce" behind the success of the recipe https://github.com/vondele/nettest/pull/329

### DIFF
--- a/model/lightning_module.py
+++ b/model/lightning_module.py
@@ -8,6 +8,7 @@ from torchmetrics import MeanMetric, MetricCollection
 from .config import NNUELightningConfig
 from .model import NNUEModel
 from .quantize import QuantizationConfig
+from .lambda_utils import LambdaController
 
 
 def _get_parameters(layers: list[nn.Module], get_biases: bool = False):
@@ -47,11 +48,6 @@ def calculate_sf_loss(scorenet, score, outcome, loss_params, actual_lambda):
 
 
 class NNUE(L.LightningModule):
-    """
-    lambda_ = 0.0 - purely based on game results
-    0.0 < lambda_ < 1.0 - interpolated score and result
-    lambda_ = 1.0 - purely based on search scores
-    """
 
     def __init__(
         self,
@@ -80,8 +76,8 @@ class NNUE(L.LightningModule):
         # lazy init so `resume_from_model` with config changes works correctly
         self.optimizer_wrapper = None
 
-        # register jitter buffer
-        self.register_buffer("jitter_buffer", torch.zeros(1), persistent=False)
+        # Initialize the lambda controller
+        self.lambda_scheduler = LambdaController(config.loss_params)
 
         self.loss_metrics = MetricCollection(
             {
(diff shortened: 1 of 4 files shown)


model/lightning_module.py
```diff
@@ -190,6 +195,17 @@ def on_test_epoch_end(self):
 
     def on_save_checkpoint(self, checkpoint):
         self.optimizer_wrapper.on_save_checkpoint(self, checkpoint)
+        checkpoint["jitter_buffer_value"] = self.jitter_buffer
```
xu-shawn: why do we set persistent=false but then explicitly code saving the variable?

TonyCongqianWang: this way it is not saved in the state dict, but in the top level ckpt dictionary instead. This means that it will be continued when using `resume_from_checkpoint` but reset when using `resume_from_model`.... somewhat complicated I know, but I am not aware of any better solution

xu-shawn: and what is the reason we need to have those different behaviors?

TonyCongqianWang: Jitter is not part of the model but part of the training

xu-shawn: hmm I guess maybe the best way then is to factor the lambda schedules into a callback (in a future PR)?

TonyCongqianWang: A lightning callback would probably be quite inefficient. But refactoring everything into it's own component should be quite trivial

model/lightning_module.py
```diff
@@ -257,6 +273,22 @@ def step_(self, batch: tuple[Tensor, ...], batch_idx, loss_type):
...
+            jitter_lambda_batch = loss_params.jitter_lambda_batch * math.sqrt(1 - loss_params.jitter_decay_lambda_batch ** 2)
+            batch_jitter_delta = jitter_lambda_batch * torch.randn_like(self.jitter_buffer)
+            self.jitter_buffer.mul_(loss_params.jitter_decay_lambda_batch).add_(batch_jitter_delta)
+            batch_jitter = self.jitter_buffer.expand_as(scorenet)
+            sample_jitter = torch.randn_like(scorenet) * loss_params.jitter_lambda_sample
+        else:
+            # During evaluation we move allocate all jitter to the sample level for better consistency.
+            batch_jitter = torch.zeros_like(scorenet)
+            sample_jitter = torch.randn_like(scorenet)
+            sample_jitter = sample_jitter * (loss_params.jitter_lambda_sample + loss_params.jitter_lambda_batch)
```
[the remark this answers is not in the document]
TonyCongqianWang: This is desired. The jitter doesn't just create perturbations for the gradient, it does fundamentally change the optimization criterium due to pow-exp > 2 and qp asym.

With a large enough batch size, the gradient is rather stable when only using per sample jitter

model/lightning_module.py
```diff
@@ -190,6 +195,17 @@ def on_test_epoch_end(self):
...
+
+    def on_load_checkpoint(self, checkpoint):
+        trainer = self.__dict__.get("_trainer", None)
+        is_resuming = (
+            trainer is not None and
+            getattr(trainer, "ckpt_path", None) is not None
+        )
+        if is_resuming:
+            if "jitter_buffer_value" in checkpoint:
+                self.jitter_buffer.copy_(checkpoint["jitter_buffer_value"])
```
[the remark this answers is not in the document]
TonyCongqianWang: Half of the suggestion is good, the other half is trash

model/lightning_module.py
TonyCongqianWang: Added it now 👍

TonyCongqianWang:
Maybe rerun copilot just in case? I think this is fairly ready to merge now. Recipe ablations also showed that this component is indeed necessary

TonyCongqianWang:
Good finds.. resolved

PR #477, opened 2026-05-15
Towards Perfect Cross Eval

TonyCongqianWang:
Improves quant config and removes hard coded values from the code.

The main goal is to improve readability and maintainability. As of currently there is a mismatch between training and inference (see #416 for more information). This patch intends to not only fix the mismatch but also to prevent mismatches in the future by inferring runtime correction factors directly from the quantization config instead of relying on manual changes across the codebase.

This allows arbitrary changes to the weight and activation quantization scales without having to make changes across the entire codebase.

As a quality of life improvement, serialize.py and quantize.py were refactored and the information density of the histogram callback was improved: The current histogram callback reports three values that are rather uninformative, since the model weights are already clipped through the callbacks. The logic was changed to use the native clipping of the model first and then clip only for type safety for serialization.

Allows more fine grained control over quant scheme (i.e. different weight scales per ls layer) and automatically calculates required adjustment factors to match inference instead of relying on hardcoding

ADDED FUNCTIONALITY:
- optional fake quantization of activations 
- optional fake quantization of weights
- extended cross check eval to compare different fake quantization configurations
- optional weight decay for just virtual params for numerical stability

BREAKING CHANGES:
- Quant scheme is changes slightly: Weight scale for l1 and output are changed to better match empirically observed distribution (64 -> 128)
- Activation scales are changed to clean power of two (255 -> 256, 127 -> 128) to eliminate the need for correction factors.
- Order of feature transformer weight segments slightly changed to allow easier serialization (clean looping over features).
- Output layer does not use special weight scale to directly account for `out_scale`, instead out scale is handled in inference code, so that quant scheme of skip value and output value match.


model/utils/serialize.py
```diff
@@ -73,21 +106,23 @@ def __init__(
         model: NNUEModel,
         description: str | None = None,
         ft_compression: str = "none",
+        verbose: bool = True,
```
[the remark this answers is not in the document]
TonyCongqianWang: > remains quite/fast 

Default behavior does not change

.github/workflows/cpu-testrun.yml
```diff
@@ -20,22 +20,28 @@ jobs:
       - name: Checkout Compatible Stockfish Repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
-          repository: official-stockfish/Stockfish
+          repository: TonyCongqianWang/Stockfish
```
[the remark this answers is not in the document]
TonyCongqianWang: WIP, will be changed once stockfish changes merged

TonyCongqianWang:
I added fake quantization for weights too.

This does add quite a bit of changes, but greatly reduces the `.ckpt` cross eval.

Maybe rerun Copilot review @vondele ?

model/utils/serialize.py
```diff
@@ -299,7 +301,8 @@ def read_leb_128_array(
             raise Exception("Unexpected end of file when reading compressed data.")
 
         res = torch.tensor(
-            decode_leb_128_array(d, reduce(operator.mul, shape, 1)), dtype=torch.float32
+            decode_leb_128_array(d, reduce(operator.mul, shape, 1)),
+            dtype=torch.float64
```
vondele: is float64 intentional

model/utils/serialize.py
```diff
@@ -323,7 +326,7 @@ def tensor(self, dtype: npt.DTypeLike, shape: Sequence[int]) -> torch.Tensor:
 
         if compression == "none":
             d = np.fromfile(self.f, dtype, reduce(operator.mul, shape, 1))
-            d = torch.from_numpy(d.astype(np.float32))
+            d = torch.from_numpy(d.astype(np.float64))
```
vondele: idem

model/utils/serialize.py
TonyCongqianWang: Yes... Since the values are int32, it is saver to load them in float64 to prevent truncation by the 24 bit mantissa of float32 to prevent any source of mismatch 

Though we cast them to float32 later and during serialization it is is also float32 before round and convert to int.. so maybe it is unnecessary after all...

model/optimizers/config.py
```diff
@@ -16,7 +16,7 @@ class OptimizerConfig(RangerLiteConfig, ScheduleFreeConfig):
     dense_weight_decay: float = 0.0
     """Weight decay to apply to the dense layer parameters."""
 
-    factorized_weight_decay: float = 0.0
+    factorized_weight_decay: float = 0.001
```
vondele: also changes the optimization parameters default?

TonyCongqianWang: reverted

model/utils/serialize.py
TonyCongqianWang: reverted

TonyCongqianWang:
Wanna copilot again @vondele ?

vondele:
from my point of view this is a good step forward and can be merged, but think @xu-shawn is best placed to have a look and merge.

TonyCongqianWang:
Always a good idea to have an extra pair of eyes and perhaps an extra llm review as well

model/model.py
```diff
@@ -78,26 +86,65 @@ def forward(
...
         b, bpsqt = torch.split(bp, self.L1, dim=1)
+
+        psqt_indices_unsq = psqt_indices.unsqueeze(dim=1)
+        wpsqt = wpsqt.gather(1, psqt_indices_unsq)
+        bpsqt = bpsqt.gather(1, psqt_indices_unsq)
+
         l0_ = (us * torch.cat([w, b], dim=1)) + (them * torch.cat([b, w], dim=1))
-        l0_ = torch.clamp(l0_, 0.0, 1.0)
+        if fake_quantize_acts:
+            pass # do not fake quantize sum of (quantized) weights
```
xu-shawn: can you make this just a comment?

model/quantize.py
```diff
@@ -1,161 +1,250 @@
...
-    return tensor.to(dtype)
-
+def _safe_convert(value: torch.Tensor, target_dtype: torch.dtype):
+    _info = torch.iinfo(target_dtype)
+    # Symmetric range: [-max, max]
+    min_val = -_info.max
+    max_val = _info.max
+
+    rounded_value = value.round()
+    clamped_value = rounded_value.clamp(min_val, max_val)
```
xu-shawn, line 24: is there any benefit in explicitly calculating the clipped values? An error is raised anyways when the values are out of range?

model/quantize.py
```diff
@@ -1,161 +1,250 @@
...
+
+    return value
 
 @dataclass
 class QuantizationConfig:
     nnue2score: float = 600.0
-    weight_scale_hidden: float = 64.0
+    weight_scale_l1: float = 128
+    weight_scale_l2: float = 64.0
+    # weight_scale_l_out = (self.nnue2score * self.weight_scale_out) / self.hidden_quantized_one
```
xu-shawn: is this correct? 600 * 16 / 128 = 75

xu-shawn: if the variable depends on other fields we should probably make it a `@property`

model/quantize.py
```diff
@@ -1,161 +1,250 @@
...
         self,
-        bias: torch.Tensor,
-        weight: torch.Tensor,
-        psqt_weight: torch.Tensor,
-    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
-        bias = bias.divide(self.ft_quantized_one)
-        weight = weight.divide(self.ft_quantized_one)
-        psqt_weight = psqt_weight.divide(self.nnue2score * self.weight_scale_out)
+        bias: Optional[torch.Tensor],
+        weight: Optional[torch.Tensor],
```
xu-shawn: Why make the parameters optional? Are we passing in `None` anywhere?

model/quantize.py
```diff
@@ -1,161 +1,250 @@
...
-        psqt_weight: torch.Tensor,
-        callback: Callable = lambda *args, **kwargs: None,
-    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
-        bias = safe_convert(bias.mul(self.ft_quantized_one).round(), torch.int16)
-        weight = safe_convert(weight.mul(self.ft_quantized_one).round(), torch.int16)
-        psqt_weight = safe_convert(
-            psqt_weight.mul(self.nnue2score * self.weight_scale_out).round(),
-            torch.int32,
-        )
+        bias: Optional[torch.Tensor],
```
xu-shawn: Can we instead split this into `quantize_feature_transformer_bias(self, bias, callback)` and `quantize_feature_transformer_weight(self, weight_psqt_weight, export_dtype)` to reduce some clutter

cross_check_eval.py
```diff
@@ -135,7 +137,7 @@ def compute_basic_eval_stats(evals):
     return min_val, max_val, avg_val, avg_abs_val
 
 
-def compute_correlation(engine_evals, model_evals, fens):
+def compute_correlation(engine_evals, model_evals, fens, title="CROSS-CHECK EVALUATION SUMMARY", sf_name="Py", py_name="SF"):
```
xu-shawn: why are the names swapped? the default arguments are confusing

model/modules/features/input_feature.py
```diff
@@ -20,6 +20,9 @@ def merged_weight(self) -> torch.Tensor:
     @abstractmethod
     def coalesce(self) -> None: ...
 
+    @abstractmethod
+    def zero_virtual_weights(self) -> torch.Tensor: ...
```
xu-shawn: should be `... -> None`?

cross_check_eval.py
```diff
@@ -194,10 +197,18 @@ def compute_correlation(engine_evals, model_evals, fens):
 
     # 1. Values Summary
     print(
-        f"{'Average Absolute Value':<30} | {en_abs_avg:>38.2f} | {sum(d['q_py'] for d in data) / len(data):>38.4f}"
+        f"{f'Average Absolute Value ({sf_name})':<30} | {en_abs_avg:>38.2f} | {sum(d['q_sf'] for d in data) / len(data):>38.4f}"
     )
     print(
-        f"{'Min / Max Value':<30} | {en_min:>17.1f} / {en_max:<18.1f} | {min(d['q_py'] for d in data):>17.4f} / {max(d['q_py'] for d in data):<18.4f}"
+        f"{f'Min / Max Value ({sf_name})':<30} | {en_min:>17.1f} / {en_max:<18.1f} | {min(d['q_sf'] for d in data):>17.4f} / {max(d['q_py'] for d in data):<18.4f}"
```
xu-shawn: shouldn't the max value also use `q_sf`?

vondele:
FYI, I'm also testing training once more with the current branch here https://github.com/vondele/nettest/pull/350

cross_check_eval.py
TonyCongqianWang: Indeed... Do you have some naming recommendations?

model/quantize.py
TonyCongqianWang: Good idea..

model/quantize.py
TonyCongqianWang: Ah it used to be necessary but isn't anymore..

model/quantize.py
TonyCongqianWang: The comment outdated. I will remove it

model/model.py
TonyCongqianWang: Sure .. functionally the same. Though I find this easier to read

model/quantize.py
TonyCongqianWang: It gives the number of OOR values in the debug fwiw.. Small but tangible benefit imo

ftperm.py
```diff
@@ -537,45 +535,17 @@ def make_sparse_batch_provider(
...
-
-    l0_s = torch.split(l0_, model.L1 // 2, dim=1)
-    l0_s1 = [l0_s[0] * l0_s[1], l0_s[2] * l0_s[3]]
-    # We multiply by 255/512 because in the quantized network 1.0 is represented by 255
-    # and we want to scale to 1.0=127, but a shift is faster than a division (in inference)
-    l0_ = torch.cat(l0_s1, dim=1) * (1 / 512)
-
-    # Inference uses bitshift which is equivalent to rounding down (floor).
-    return l0_.floor()
+        f.weight.data = f.weight.data.div_(model.quantization.ft_quantized_one)
```
43eayl5: Directly assigning or modifying `.data` is a legacy PyTorch anti-pattern.
Should use `with torch.no_grad():` blocks paired with in-place operations or `copy_`.

cross_check_eval.py
```diff
@@ -288,27 +299,33 @@ def main():
...
+    nnue = read_model(
+        cross_check_config.net,
+        config=nnue_lightning_config,
+    )
+    nnue.to(cross_check_config.device)
+    nnue.eval()
     # --net - returns the NNUEModel directly
-    inner_model = model.model if isinstance(model, M.NNUE) else model
-    input_feature_name = inner_model.input_feature_name
+    nnue_model = nnue.model if isinstance(nnue, M.NNUE) else nnue
```
43eayl5: `reader.model` returns an instance of `NNUEModel`. Is this check necessary?

model/modules/stacked_linear.py
```diff
@@ -22,8 +36,16 @@ def _init_uniformly(self) -> None:
...
 
-    def forward(self, x: torch.Tensor, ls_indices: torch.Tensor) -> torch.Tensor:
-        stacked_output = self.linear(x)
+    def forward(self, x: torch.Tensor, ls_indices: torch.Tensor, fake_quantize_weights: bool=False) -> torch.Tensor:
+        weight = self.linear.weight
+        bias = self.linear.bias
+        if fake_quantize_weights:
+            if self.quantization is None or self.layer_key is None:
+                raise RuntimeError("self.quantization and self.layer_key are required to use fake quantize weights.")
+            weight = self.quantization.fake_quantize_weights(weight, f"{self.layer_key}_weight")
```
43eayl5, line 45: The current implementation of `QuantizationManager` introduces a dictionary, mapping dynamic string keys using string interpolation. This requires passing a `layer_key: str` parameter to `StackedLinear` and a whole `QuantizationManager` instance. It also introduces dynamic lookups in hot paths (`forward` during training) and fragile `RuntimeError` branches if these fields are omitted or mismatch.
You can make the weight and bias scales first-class attributes of the stacked linear layers. Passing the concrete numeric scales directly to the constructors of `StackedLinear` and `FactorizedStackedLinear` decouples them from `QuantizationManager`, makes the parameters explicitly typed and eliminates string interpolation and potential lookup failures.

ftperm.py
TonyCongqianWang: Thats correct, just a vestige from the past.. can change

cross_check_eval.py
TonyCongqianWang: Ckpt does not

model/modules/stacked_linear.py
TonyCongqianWang: This isn't really an issue due to torchs compile

TonyCongqianWang: Having all of the quantization logiv inside the quantization manager is somewhat of an old design choice that I adhered to.

Could be changed yes, but that would need some more discussion. I'd rather not randomly start putting quant scales inside the layers though it's a possibility if done consistently

cross_check_eval.py
43eayl5: Makes sense... since loading a checkpoint returns the Lightning wrapper. Just to double-check:
For `nnue`: `read_model` always returns `NNUEModel`, so won't `isinstance(nnue, M.NNUE)` always be false?
For `ckpt`: `M.NNUE.load_from_checkpoint` always returns an `M.NNUE` wrapper, so won't it always be true?

TonyCongqianWang: Hmm yes. That's correct. Could be refactored to ge more direct. It's always a tradeoff between cleaning up the code and not blowing up the diff too much, but since the code is touched anyway, I can do it

vondele:
So nettest run at b4fd490092544a8cc6583fcd0d404383ba570891 completed and looks good.

TonyCongqianWang:
I think I got everything now! Thanks for the reviews

xu-shawn (approved):

ftperm.py
```diff
@@ -537,45 +535,18 @@ def make_sparse_batch_provider(
...
-
-    l0_s = torch.split(l0_, model.L1 // 2, dim=1)
-    l0_s1 = [l0_s[0] * l0_s[1], l0_s[2] * l0_s[3]]
-    # We multiply by 255/512 because in the quantized network 1.0 is represented by 255
-    # and we want to scale to 1.0=127, but a shift is faster than a division (in inference)
-    l0_ = torch.cat(l0_s1, dim=1) * (1 / 512)
-
-    # Inference uses bitshift which is equivalent to rounding down (floor).
-    return l0_.floor()
+        f.weight.mul(model.quantization.ft_quantized_one).round_()
```
xu-shawn: Claude flagged this. `tensor.mul` creates a new tensor, so this line is a no-op

xu-shawn:
Here's the rest of the outputs from Claude:

 ### 🟠 Medium — behavior change worth confirming

  **In-place mutation of autograd tensors in `ComposedFeatureTransformer.forward` (`composed.py:69-72`)**

  ```python
  merged[:, :self.l1_size]  = self.quantization.fake_quantize_weights(merged[:, :self.l1_size], "ft_weight")
  merged[:, self.l1_size:]  = self.quantization.fake_quantize_weights(merged[:, self.l1_size:], "ft_psqt_weight")
  bias = self.quantization.fake_quantize_weights(bias, "ft_bias")
  bias[self.l1_size:].zero_()
  ```

  This works (CI trains with both flags defaulting `True`), but mutating a grad-requiring `merged`/`bias` in place is fragile — easy to break with a future autograd-sensitive
  change, and harder to reason about than building a new tensor. Consider:
  ```python
  w  = self.quantization.fake_quantize_weights(merged[:, :self.l1_size], "ft_weight")
  pw = self.quantization.fake_quantize_weights(merged[:, self.l1_size:], "ft_psqt_weight")
  merged = torch.cat([w, pw], dim=1)
  ```
  Also note the `bias[self.l1_size:].zero_()` is **redundant** — `init_weights` already zeros the PSQT bias (`composed.py:106`) and it receives no gradient (the PSQT bias
  cancels in `(wpsqt - bpsqt)`), so this only runs in the fake-quant path and is a no-op in effect. Either drop it or document why it's defensive.

  ---

  ### 🟡 Low — quality / clarity

  - **`cross_check_eval.py:166`** — zero-engine-eval `rel_err` denominator changed `1.0 → 0.001`, a 1000× inflation that will dominate the "TOP 5 LARGEST RELATIVE ERRORS"
  table whenever `e == 0`. If intentional (to surface zero-eval positions), fine; otherwise it makes that table misleading.

  - **`compute_correlation` internal naming** — parameters are now generic `cmp_evals/ref_evals`, but the internal dict still hardcodes `"sf"`/`"py"`/`q_sf`/`q_py`. For the 
  `"CKPT VS NNUE"` call the column labeled with `ref_name` actually shows NNUE values stored under key `"sf"`. The printed output is labeled correctly, but the variable 
  naming is now a trap for the next editor.

  - **FT serialization layout** — weights/psqt are now interleaved **per feature** (`segment_weight` then `segment_psqt_weight`) instead of master's "all weights then one 
  psqt block". Read/write are symmetric and it's byte-identical for single-feature nets (HalfKAv2_hm), so standard nets are unaffected; just confirm any external consumer of 
  multi-feature (threats) nets agrees on the new layout.

  - **`fake_quantize_skip_act` (`quantize.py`) is a pure identity** wrapped in an `if fake_quantize_acts:` guard in `layer_stacks.py` — dead indirection unless it's a 
  deliberate placeholder; a comment would help.

  - **Type annotations**: `weight_scale_l1: float = 128` and `weight_scale_l_out: float = 128` assign `int` literals to `float`-annotated fields (cosmetic; prefer `128.0`).

  - **Default mismatch**: `NNUEModel.forward` defaults `fake_quantize_*=True` while `LayerStacks.forward` defaults them `False`. Harmless today (the model always forwards 
  explicit values), but the divergence is a footgun if either is ever called directly.

ftperm.py
TonyCongqianWang: oh yes.. good catch. needs to be `mul_`

TonyCongqianWang:
Should be fixed now. Let me watch the CI to confirm...

TonyCongqianWang:
There are some values that do look a bit different after changing to in place operations... gemini told me that it is safer to not chain them. Maybe it will then be equivalent to the results when it was operating directly on the data.

TonyCongqianWang:
I am a bit confused... Somehow ftperm seems to be able to affect crosscheckeval performance... I don't understand why. Can someone check? Also it's probably better to remove quantize_ft from ft_perm entirely as weight quantization is now supported by the model itself

vondele:
Which test did you to that showed this ft_perm effect, might make it easier to reproduce for others?

TonyCongqianWang:
My last few commits only targeted ftperm and somehow affected the ckpt vs sf and nnue vs sf crosseval errors (mean abs error and also the percentiles). Could also be just random noise....

vondele:
Naively, I would expect that ftperm just acts on the integers and the result should be invariant? Could something be hitting an overflow?

TonyCongqianWang:
> Naively, I would expect that ftperm just acts on the integers and the result should be invariant? 


Yes that's also my assumption so that's why I am confused.


Though the ckpt is also different... So maybe check whether both versions produce the same crosscheck results (with the same crosscheck version)

TonyCongqianWang:
I removed the ``quantize_ft`` function from ft_perm now as it was already redundant anyway. But maybe confirm that ft_perm does about the same across the different versions (e.g. by only restoring `ft_perm` and keeping the checkpoint and cross_check_eval).

I can't do that right now unfortunately, but fwiw the CI numbers do "look right" now... not sure why there was a mean average error of 10 before, which seemed very off

vondele:
Are you sure you saw differences, I checked rather carefully the branch, and all is as expected for me, in particular both bench and cross_check are identical after ft_perm.

I ran this:
```
python -u serialize.py /workspace/scratch/17a59d9c1ceb/run/lightning_logs/version_0/checkpoints/last.ckpt nonopt.nnue --features=Full_Threats+HalfKAv2_hm^ --ft-compression=leb128 --l1=1024 --l2=31 
python -u serialize.py nonopt.nnue last.nnue --ft_optimize_data=/workspace/data/official-stockfish/master-binpacks/fishpack32.binpack --device=0 --features=Full_Threats+HalfKAv2_hm^ --ft-compression=leb128 --ft-optimize --ft-optimize-count=100000 --l1=1024 --l2=31
python -u serialize.py /workspace/scratch/17a59d9c1ceb/run/lightning_logs/version_0/checkpoints/last.ckpt try.nnue --ft_optimize_data=/workspace/data/official-stockfish/master-binpacks/fishpack32.binpack --device=0 --features=Full_Threats+HalfKAv2_hm^ --ft-compression=leb128 --ft-optimize --ft-optimize-count=100000 --l1=1024 --l2=31
printf "setoption name EvalFile value nonopt.nnue\nbench\n" | /workspace/scratch/packages/stockfish/b1053e60b7e0fe3187006c4d7f38dbd22eaff763-profile-build/Stockfish/src/stockfish 2>&1 | grep "Nodes searched"
printf "setoption name EvalFile value last.nnue\nbench\n" | /workspace/scratch/packages/stockfish/b1053e60b7e0fe3187006c4d7f38dbd22eaff763-profile-build/Stockfish/src/stockfish 2>&1 | grep "Nodes searched"
printf "setoption name EvalFile value try.nnue\nbench\n" | /workspace/scratch/packages/stockfish/b1053e60b7e0fe3187006c4d7f38dbd22eaff763-profile-build/Stockfish/src/stockfish 2>&1 | grep "Nodes searched"
python -u cross_check_eval.py --engine /workspace/scratch/packages/stockfish/b1053e60b7e0fe3187006c4d7f38dbd22eaff763-profile-build/Stockfish/src/stockfish --data /workspace/data/official-stockfish/master-binpacks/fishpack32.binpack --device=cuda --count=20000 --features=Full_Threats+HalfKAv2_hm^ --l1=1024 --l2=31 --net last.nnue > last.cc
python -u cross_check_eval.py --engine /workspace/scratch/packages/stockfish/b1053e60b7e0fe3187006c4d7f38dbd22eaff763-profile-build/Stockfish/src/stockfish --data /workspace/data/official-stockfish/master-binpacks/fishpack32.binpack --device=cuda --count=20000 --features=Full_Threats+HalfKAv2_hm^ --l1=1024 --l2=31 --net nonopt.nnue > nonopt.cc
python -u cross_check_eval.py --engine /workspace/scratch/packages/stockfish/b1053e60b7e0fe3187006c4d7f38dbd22eaff763-profile-build/Stockfish/src/stockfish --data /workspace/data/official-stockfish/master-binpacks/fishpack32.binpack --device=cuda --count=20000 --features=Full_Threats+HalfKAv2_hm^ --l1=1024 --l2=31 --net try.nnue > try.cc
diff last.cc nonopt.cc 
diff last.cc try.cc 
```
all looks good.

@xu-shawn can be merged from my point of view.

vondele (approved):

TonyCongqianWang:
Great news!

> Are you sure you saw differences

I didn't look to closely... Just saw the values in the CI, which are heavily affected by training run randomness

vondele:
Ah, of course, from training run to training run, things might be quite different.

PR #478, opened 2026-06-02
[WIP] Better skip connections

TonyCongqianWang:
The last net update caused the net to not be able to output huge values anymore, which led to a regression in mating ability. This patch aims to tackle this weakness by adding more skip connections


TonyCongqianWang:
Seems to be worse than #480

PR #479, opened 2026-06-05
Using custom lightweight modelcheckpoint implementation

TonyCongqianWang:
# Motivation

Native Lightning Callback `ModelCheckpoint` is very bloated and has one crucial limitation for our use-case:
- There is no (clean) mechanism to pause checkpointing during some periods.

This is a problem because no resuming is supported during SWA epochs. This means that checkpoints created during that time are non-functional and can overwrite crucial save-points leading to corrupted runs that have to be restarted.

This implementation handles the pause of checkpointing during SWA epochs.

# Other Benefits

Since we only use a small fraction of the bloated features offered by `ModelCheckpoint` this implementation is a lightweight alternative that is easier to maintain as `ModelCheckpoint` is poorly documented and has behavior that is difficult to understand. This lightweight implementation on the other hand is easy to parse and any behavior can be quickly deduced. If `ModelCheckpoint` does something weird, it takes A LOT more effort to debug!

Mainly:
- `ModelCheckpoint` does NOT automatically save a checkpoint at the end of training -- even when using `save-last` which is very counter-intuitive. This implementation always saves a checkpoint at the end of training -- without the need of any subclassing!
- `ModelCheckpoint`'s checkpoint frequency is depending on whether we use validation and on how often we validate! This is because its use is mainly tied to Monitor to keep the *best* checkpoints. Since we don't use any Monitor, this doesn't make any sense to us. We want checkpoints to be saved in a predictable frequency. As such this lightweight implementation saves checkpoints always at the save frequency.

### DIFF
--- a/model/__init__.py
+++ b/model/__init__.py
@@ -1,4 +1,4 @@
-from .callbacks import WeightClippingCallback, ExplicitSWACallback
+from .callbacks import WeightClippingCallback, ExplicitSWACallback, SimplePeriodicCheckpoint
 from .config import ModelConfig, LossParams, NNUELightningConfig
 from .optimizers import OptimizerConfig, RangerLiteWrapper, ScheduleFreeWrapper
 
@@ -22,6 +22,7 @@ from .utils import (
 __all__ = [
     "WeightClippingCallback",
     "ExplicitSWACallback",
+    "SimplePeriodicCheckpoint",
     "ModelConfig",
     "LossParams",
     "add_feature_args",
--- a/model/callbacks.py
+++ b/model/callbacks.py
@@ -158,6 +158,8 @@ class SimplePeriodicCheckpoint(L.Callback):
         if not self.save_last:
             assert self.every_n_epochs >= 1, "every_n_epochs must be at least 1 when save_last is False."
 
+    @torch.no_grad()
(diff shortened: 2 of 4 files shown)


TonyCongqianWang:
Can you do some testing on your end?

@vondele @Disservin

xu-shawn:
Could you elaborate on your justifications for a +100 line diff? I'm not sure why we need it or if it could be done without adding an entire callback

TonyCongqianWang:
> Could you elaborate on your justifications for a +100 line diff? I'm not sure why we need it or if it could be done without adding an entire callback

No it can't be done without adding an entire callback because it replaces an entire callback

xu-shawn:
That doesn't quite answer my concern... What's the motivation behind replacing the callback to begin with?

TonyCongqianWang:
> That doesn't quite answer my concern... What's the motivation behind replacing the callback to begin with?

Ah okay I understand. I will write more about it in the description. I have talked to vondele a bit so the context is missing here. It is mainly for recipe simplification that aren't possible due to limitations of the native callback.

Frankly, the native callbacks in Lightning all kinda suck

TonyCongqianWang:
I updated the text

PR #482, opened 2026-06-10
[WIP] Optimize data loader with C++ pinned memory allocations

TonyCongqianWang:
WIP...

PR #483, opened 2026-06-12
[WIP] Optimize data loader and eliminate redundant feature value tensors

TonyCongqianWang:
- Implement C++ pinned memory allocations and contiguous grouping for faster H2D transfers.
- Remove redundant 1.0f value tensors (white_values and black_values) from the dataloader, PyTorch models, and Cupy kernels.
- Update downstream training, evaluation, permutation, and sigmoid fitting scripts to unpack and use 8 tensors instead of 10.
- Update unit tests for device parity and feature transformers to align with the new signatures.


TonyCongqianWang:
Superseeds #482

TonyCongqianWang:
Ah.. the branch was based of the wrong commit..

PR #487, opened 2026-06-23
Add pytest to requirements and ci

TonyCongqianWang:


TonyCongqianWang:
@vondele @xu-shawn please review. I don't see a reason to not merge this.

The longer we postpone this, the higher the risk of having broken/outdated pytests.

We should strive to keep the repository as healthy as possible

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,32 @@
...
+    types: [ opened, synchronize, reopened, ready_for_review ]
+  push:
+    branches: [master]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  ruff:
```
xu-shawn: typo?

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,32 @@
...
+          python-version: "3.12"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+
+      - name: Run pytest
+        run: |
+          pytest tests/test_feature_transformer.py tests/test_sparse_linear_device_parity.py
```
xu-shawn: `tests/test_sparse_linear_device_parity.py` doesn't do anything on ubuntu-latest. maybe just drop it here?

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,32 @@
...
+  ruff:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Set up Python
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: "3.12"
```
xu-shawn: claude suggests `cache: 'pip'` here since we have heavy dependencies. could you check if it's viable?

.github/workflows/pytest.yml
TonyCongqianWang: Hmm I thought so too, but included it for the sake of completeness, since it's not guaranteed to stay that way..

Probably it's better to just call pytests from the root, now that I think about it... That was we won't forget anything

.github/workflows/pytest.yml
TonyCongqianWang: Good idea. I'll investigate

TonyCongqianWang:
Should be fixed now..

TonyCongqianWang:
Instead of caching pip I added docker layer caching. Should be shared between both jobs hopefully

TonyCongqianWang:
This should increase the overall long term health of the repository. Chached docker containers indeed seem to work and speed up ci runs substantially

PR #493, opened 2026-06-23
Add Dataloader Chunkreader

Disservin:
Reorganizes the logic into a chunkreader to avoid duplicating the internals in the parallel dataloader implementation.


xu-shawn:
needs rebase

PR #408, opened 2026-02-26
Composable Inputs

xu-shawn:
Restore feature composition support, which was present but nonfunctional before its removal in #397. Following this, the psq weights are extracted out of FullThreats. The current SF inputs are refactored into a composed feature of `Full_Threats` and `HalfKAv2_hm^`.

This PR breaks compatibility in two ways:

1. New network hash: engine-side updates can be found at https://github.com/xu-shawn/Stockfish/tree/new_net_hash.
2. New feature name. `--features="Full_Threats^"` is replaced by `--features="Full_Threats+HalfKAv2_hm^"`

I've verified serialize and dataloader and both seems correct. Need to measure training speed since there's now an extra tensor concatentation when forwarding inputs.

Marking as draft since ftperm needs fixing

PR #494, opened 2026-06-24
Refactor feature transformer and add fused kernel

TonyCongqianWang:
reimplementation of #451:

fusing forward_ft into one kernel (except the fake quantization)


TonyCongqianWang:
@vondele @Disservin @xu-shawn 

please test

vondele:
noticeable speedup

```
master 65536 1 68.12it/s
double_ft_fused 65536 1 73.89it/s
master 65536 2 70.37it/s
double_ft_fused 65536 2 73.55it/s
master 65536 3 70.46it/s
double_ft_fused 65536 3 74.25it/s
```

TonyCongqianWang:
Nice. lets merge.

Ill ask gemini tomorrow if there is more to fuse.

Could you perhaps profile to see what is actually a current bottleneck?

Also can you test at doubled batch size to see the scaling behavior?

Would help to find a direction on where to optimize..

model/model.py
```diff
@@ -5,6 +5,9 @@
 from .modules import LayerStacks, ComposedFeatureTransformer, get_feature_cls
 from .quantize import QuantizationManager
 
+from .modules.feature_transformer.functions import FusedDoubleFtFunction, _HAS_CUPY_KERNELS
+
+USE_FUSED_DOUBLE_FT = True
```
xu-shawn: can this be removed?

TonyCongqianWang: No, it is required for the pytest.

An alternative would be to pass it to the foward_ft function itself defaulting to true

Imo it is essential to keep both branches in case we want to change the architecture at some point and keeping the pytest is essential to guarantee it is matching.

If you have another suggestion to fulfill the requirements, feel free to share

xu-shawn: where in the testing code is it used? I think any solution that doesn't require a global variable here is fine

TonyCongqianWang: Huh... in this version it is actually not used... Here the code is duplicated at

https://github.com/official-stockfish/nnue-pytorch/pull/494/changes#diff-5999b5f77aeca29720b42fe6a731d18dc9c7a39e05150968c73ebdd6fa06b1f8R61

which should be fixed.

I do somewhat dislike the approach of adding a parameter since it is not meant to be disabled in normal circumstances.

A global variable in python isn't the same as in other languages, as it is tied to the file and can be toggled from external code.

I am not really sure what the best approach is. So feel free to make a patch

TonyCongqianWang: Hmm perhaps the cleanest solution is to remove the deprecated BaseFeatureTransformer and DoubleFeatureTransformer classes. change composed features to NOT have a forward an instead only to give the weights. And the create a new DoubleFeatureTransformer class that uses composition:
it has a composed input feature as member and its forward function is essentially the forward_ft in the current model.py.

But as for how to control which branch to take... I am still not sure what the best approach is. If you look at libraries like huggingface, they absolutely do use "global" variables to control things like using fused spda or which spda backend to use. So it is a "pythonic" way to do things

TonyCongqianWang:
@vondele @xu-shawn please re-review

Doing a cleanup does bloat the diff unfortunately, but I would say this is a lot cleaner now. It does not contain the deprecated FeatureTransformer implementations and the composed is actually a thin wrapper as it claims and doesn't smuggle some forward logic

TonyCongqianWang:
Confirmed to work
https://github.com/vondele/nettest/pull/380

Pytest and bench script also working

Please work

model/modules/feature_transformer/functions.py
```diff
@@ -149,3 +154,147 @@ def apply(feature_indices, weight, bias):
...
+        )
+
+        return None, None, None, None, None, grad_weight, grad_bias, None, None
+
+
+class FusedDoubleFtFunction:
+    @staticmethod
+    def apply(us, them, white_indices, black_indices, psqt_indices, weight, bias, max_ft_activation, l1_size):
+        if not _HAS_CUPY_KERNELS:
+            raise RuntimeError("CuPy kernels not available for FusedDoubleFtFunction")
```
xu-shawn: Can we have a pytorch fallback here as well?

TonyCongqianWang: Not sure if that's the correct pattern to use

The fused branch should only be called when cuda is available.

Allowing a fallback here would allow silent errors

xu-shawn: I was thinking of changing DoubleFeatureTransformer as well to always use FusedDoubleFtFunction. This is more consistent with how SparseLinearFunction is currently written

TonyCongqianWang: I don't think that's a good idea... We shouldn't lock ourselves into one architecture like that.


We should keep the separated path to allow for future innovation

xu-shawn: How exactly does that lock us in one architecture?

TonyCongqianWang: With the fused kernel, intermediate results become inaccessible. 

I'll give you concrete examples that become 
Impossible without having to change the cuda kernel:

- regularization of the pre activation value or pre pointwise multiply value.

- change or crelu to sqr crelu. Or any change of the activation function in general.

- implementation of mixed scale activation quantization

I'm sure other people easily can come up with more ideas that involve require intermediate calculations. If we remove the fallback branch, we  make development of those ideas prohibitively difficult

xu-shawn: I'm not suggesting that we remove the fallback branch entirely, just move them into functions.py so that we 1) don't duplicate cuda arch detection and 2) maintain consistency with the sparse linear function

TonyCongqianWang: Ah I see. I misunderstood. Maybe you can supply a patch then?

I aggree that duplicate fallback detection is undesired.

I also think that it's strictly better to keep the fallback as well as the fused call at the same place for maintainability.

How to switch between the different fallbacks is something you can decide. There are many different ways to do that, though there is a distinction -- one of them is truly a fallback, the other one is a deliberate decision: if there is cupy available, then both sparse linear as well as fused are possible.

If looking at torch, a common pattern might be:

Have a "set_impl" function that accepts a string that represents an enum. And one can choose between the 3 different impls. The default is then decided by whether cupy is available (use fused double ft) or no (use vanilla torch fallback).

This way we keep the in-between fallback for maintainability without doubling the detection

TonyCongqianWang: I updated accordingly

TonyCongqianWang: I'll resolve this comment. Feel free to write a new one in the new version

TonyCongqianWang:
Resolved everything. If the ci runs smoothly I think its good to go?

TonyCongqianWang:
Could you please take a look again @xu-shawn. Getting this merged would mean a decent speedup for out training runs. With the amount of training runs we are doing right now, it is quite substantial in total

model/modules/feature_transformer/module.py
```diff
@@ -1,67 +1,78 @@
...
 
 
-class BaseFeatureTransformer(nn.Module):
-    def __init__(self, num_inputs, num_outputs):
+class DoubleFeatureTransformer(nn.Module):
+    def __init__(self, features: ComposedFeatures):
         super().__init__()
-        self.num_inputs = num_inputs
-        self.num_outputs = num_outputs
+        self.features = features
```
xu-shawn: what's the reason that DoubleFeatureTransformer now holds a feature?

model/modules/feature_transformer/functions.py
```diff
@@ -137,15 +141,211 @@ def backward(ctx, grad_output):
...
+# Allowed modes: "auto", "fused", "sparse", "torch"
+_DOUBLE_FT_IMPL = "auto"
+
+
+def set_double_ft_impl(mode: str):
+    """Set the implementation mode for the double feature transformer.
+
+    Allowed modes: "auto", "fused", "sparse", "torch"
+    """
+    global _DOUBLE_FT_IMPL
```
xu-shawn: I'm not a fan of this global system. what's preventing us from passing in a backend argument like `SparseLinearFunction`?

TonyCongqianWang: I don't understand what you are saying. Please clarify

model/modules/feature_transformer/module.py
TonyCongqianWang: Thats something you implemented back in 
https://github.com/official-stockfish/nnue-pytorch/pull/408

Here in particular

https://github.com/official-stockfish/nnue-pytorch/pull/408/changes#diff-e914f64d3990e0177b12289df868c9031efa861e94889a4421406c5449ee70dfR8

model/modules/feature_transformer/functions.py
xu-shawn: I don't like that _DOUBLE_FT_IMPL is a global parameter configured by a function, wher a call to `set_double_ft_impl` now changes the backend of all instantiated `DoubleFeatureTransformer`. I'd prefer that it is passed into DoubleFeatureTransformer as an argument, mirroring what's already in place for SparseLinearFunction.

model/modules/feature_transformer/module.py
xu-shawn: I know what ComposedFeatures is, but I'm not sure why it is now a field in the DoubleFeatureTransformer class.

TonyCongqianWang: Because ComposedFeatures isn't supposed to be a feature transformer. Your description reads:


```
    """Thin coordinator that wraps one or more InputFeature modules.

    Each feature owns its own weight parameters. This class owns the shared
    bias and delegates everything else to the underlying features.
    """

```

The double feature transformer is -- in my opinion -- the correct place for a forward function. Just like it used to be before your patch...

model/modules/feature_transformer/functions.py
TonyCongqianWang: Ahhh ok. I now understand, you want the `DoubleFeatureTransformer` to have a `backend` parameter?

Well that could be done. But I don't think it is the right way to be honest. My role model for this is how pytorch handles sdpa. They also use a global getter/setter to control the backend and I agree with them. I think it is the cleanest approach as a function parameter would suggest that it is something you can change whenever you want, when it really is only supposed to be changed for edge cases (like tests and debugging).

If you prefer, I can change it to a parameter anyway

model/modules/feature_transformer/module.py
xu-shawn: Ah I see how it is now. It's been so long that I've forgotten DoubleFeatureTransformer isn't used anymore in model....

In that case, I would prefer to remove both FeatureTransformer and DoubleFeatureTransformer, and put all the forward logic to ComposedFeatureTransformer. Currently, the only job I see DoubleFeatureTransformer doing is acting as a wrapper to implement forward for ComposedFeatures, and in doing so every operation to features now crosses encapsulation with an awkward `self.input.features`.

If we are refactoring though, I think these would be the direction I would go instead:
* Scrap the comment that "everything else" is delegated to the underlying features. imo some operations, like forward, can really only be implemented effectively by the composition as a whole
* Keep the `merged_weight_and_bias` thing. I think this indeed deserves a separate function
* Since ComposedFeaturesTransformer is doing more than just a list of features with convenience functions, move ComposedFeaturesTransformer to modules/feature_transformer and move the features module to be under feature_transformer

In any case I would prefer (if possible) that refactors go in another PR, such that it can be separately reviewed and worked on.

How does that sound?

TonyCongqianWang: Ok that sounds fine. Though I do think there is some merit to keeping both classes separate as that increases maintainability in my opinion. Especially now that the forward method isn't trivial anymore.

FeatureTransformer and DoubleFeatureTransformer are indeed dead code in the current master which is why removing them also makes sense.

I could rename DoubleFeatureTransformer to ComposedFeaturesTransformer as it does take a ComposedFeature

That way we at least separate forward logic from weight merging logic which I think is a reasonable thing to do? Especially since with this refactor ComposedFeatures would actually share the interface with features, which I think has some merit of its own in terms of clean abstractions

> in doing so every operation to features now crosses encapsulation with an awkward self.input.features

this is a valid concern. I didn't find it all that awkward, but now I see where you are coming from. Instead we could also promote self.features to a first class member and feed merged_weight and bias to inputs forward method and input becomes purely functional?

My biggest concern is that we would otherwise have one very bloated class.

xu-shawn: Right, I think we probably should keep pairwise mul logic out of Composed.

What I was saying about encapsulation are things like `self.input.features.init_weights()` where the necessary methods aren't directly visible in self.inputs, and therefore we have to break encapsulation just to initialize self.inputs.

We can probably expose these methods too in `DoubleFeatureTransformer`, but I think it's better for FT management and the fused operations be delineated.

What I am now thinking, is that all the code under the `else` branch be moved to `FusedDoubleFtFunction`. We call the `FusedDoubleFtFunction.apply()` in forward and FusedDoubleFtFunction decides whether to route to the cuda or pytorch implementations

TonyCongqianWang:
Alright, I think I did everything you asked for @xu-shawn. Please take a look.

Note: I do not have a local GPU available so I couldn't test whether it works or not...

TonyCongqianWang:
I'd like to make it clear I also prefer refactoring to be in their own PR.

At the same time you can't have your cake and it it too.

You would have to accept some leaky abstractions like with the initial PR.

However, I do think you made some great suggestions, all points where valid and we had a productive discussion which at the end lead to a higher quality PR.

If we would merge refactors more often and faster that would also help I suppose.

At this point this patch could be seen as mostly a refactor and the fused implementation is just a cherry on top

vondele:
Still with measurable speedup:

```
master 131072 2 53.11it/s
double_ft_fused 131072 2 54.92it/s
master 131072 3 53.36it/s
double_ft_fused 131072 3 54.68it/s
```

TonyCongqianWang:
> Still with measurable speedup:
> 
> ```
> master 131072 2 53.11it/s
> double_ft_fused 131072 2 54.92it/s
> master 131072 3 53.36it/s
> double_ft_fused 131072 3 54.68it/s
> ```

This is for 4 cards right? I wonder what the speed for 1-3 cards is. Perhaps we can get a feel for how large the inter card communication overhead is?

vondele:
One GPU:
```
master 131072 1 25.30it/s
double_ft_fused 131072 1 27.68it/s
master 131072 2 26.00it/s
double_ft_fused 131072 2 27.33it/s
master 131072 3 27.04it/s
double_ft_fused 131072 3 27.78it/s
```

TonyCongqianWang:
Looks like a larger speedup in relative terms and also reveals that the inter card communication might indeed be quite a bottleneck

TonyCongqianWang:
We could try multiple local sgd steps between syncing but that's quite difficult to implement with lightning unfortunately

vondele:
> We could try multiple local sgd steps between syncing but that's quite difficult to implement with lightning unfortunately

I somehow prefer the same optimizer behavior independent of the number of GPUs used.

PR #496, opened 2026-06-27
Use load_from_checkpoint for resume_from_model

TumCucTom:
## Summary

Implements the approach discussed in #438: replace the brittle `resume_from_model` logic with PyTorch Lightning's `load_from_checkpoint`, passing the run-specific arguments explicitly.

### What it replaces

Previously, `train.py` resumed a model by loading a `.pt` with `torch.load(..., weights_only=False)` and then **overwriting** attributes (`max_epoch`, `num_batches_per_epoch`, `config`, `param_index`) on the already-constructed module. As noted in the issue, this is brittle: hyperparameters were effectively garbage from `serialize.py` and then patched again after load, which caused pain during the config refactor.

### What it does now

- `resume_from_model` now uses `M.NNUE.load_from_checkpoint(...)` (consistent with `serialize.py`, `cross_check_eval.py` and `model/utils/load_model.py`), passing `config`, `max_epoch`, `num_batches_per_epoch` and `param_index` directly to the constructor. The module is built correctly up front instead of being patched after the fact.
- Added `self.save_hyperparameters("num_psqt_buckets", "num_ls_buckets")` in the `NNUE` module so the architecture-defining hyperparameters are restored from the checkpoint, per the maintainer note that `save_hyperparameters` is needed to keep hyperparameters from the last run. The config dataclass is intentionally **not** pickled into the checkpoint, keeping it robust against future config refactors.
- Updated the `resume_from_model` docstring to reflect that it now takes a `.ckpt`.

Closes #438

### DIFF
--- a/config.py
+++ b/config.py
@@ -66,7 +66,7 @@ class TrainingConfig:
     """Torch seed to use."""
 
     resume_from_model: Optional[str] = None
-    """Initializes training using the weights from the given .pt model."""
+    """Initializes training using the weights from the given .ckpt model."""
 
     resume_from_checkpoint: Optional[str] = None
     """Initializes training using a given .ckpt model."""
--- a/model/lightning_module.py
+++ b/model/lightning_module.py
@@ -58,6 +58,13 @@ class NNUE(L.LightningModule):
     ):
         super().__init__()
 
+        # Persist only the architecture-defining hyperparameters so that
+        # `load_from_checkpoint` can reconstruct the model with the correct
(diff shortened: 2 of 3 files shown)


TumCucTom:
closing this one - the load_from_checkpoint change doesn't actually work here. --resume-from-model is fed .pt files (and .nnue via easy_train), which load_from_checkpoint can't read, and it breaks the training pipeline test. so it's wrong, my mistake.

PR #480, opened 2026-06-09
Update to SFNNv15: Better Skip Architecture

TonyCongqianWang:
Changes skip architecture to allow for wider dynamic range.

Skip from L2 to out is realized by concatenting L2 and L3 for the output layer.

L1 skip connection was modified so that it simply takes the last two values of the buffer one multiplied by 1.0 the other by 2.0 to essentially keep the "free" information as we would otherwise zero pad anyway for simd

Also contains #498

PR #498, opened 2026-06-28
Add mate score remapping

TonyCongqianWang:
Remapping of DTM DTM scores from tablebase positions.

~32k scores are hard for the net to learn as they are far away from normal labels. Additionally `#100` is scored basically the same as ``#1``. A simple configurable remap fixes both.


TonyCongqianWang:
@vondele @xu-shawn Any recommendations here? I think merging this first is best, as it would take at least one thing out of the arch update PR which is otherwise too overloaded.

vondele:
since this has merge conflicts, went straight for #480

PR #501, opened 2026-07-11
add stopping on nan callback

TonyCongqianWang:


vondele:
is this stopping with an error code?

TonyCongqianWang:
No I dont think so... Ill look into it later

TonyCongqianWang:
Now it does error code 1 and I also added a pytest

PR #503, opened 2026-07-18
Rewrite NNUE documentation

sscg13:
Replaces the old NNUE documentation with a modern explanation of NNUE from first principles, building up to state of the art design.

Also adds new architecture diagrams.

Feedback is appreciated.


Sopel97:
I understand how some updates are required as important parts have changed, but this pretty much guts 90% of the information

sscg13:
My personal view is that the old documentation had some kind of reputation for being difficult to understand for non-experts, but of course getting more opinions on this is helpful.

Most of what I've removed is training side information, which I do explicitly state in the new preface.

mstembera:
Thanks for taking the time to keep the docs up to date!  Can the training side info just be refactored but still preserved?  Likely most people reading this either are or are working to become experts.

sscg13:
Yeah, I think ideally we move training info into separate documentation, so that we can also cover things like the specific datasets we use, nettest configurations, etc.

Sopel97:
optimized inference has also been completely gutted

sscg13:
Will put this on draft and make a separate PR with just the new architecture images

Sopel97:
ok, I'm not the target audience so whatever, if there's a serious reason this is being rewritten so be it, but I'd like to see at least the following bits preserved/rewritten/updated

- quantization derivation
- at least basic optimized SIMD code with diagrams for complex operations
- layer types that are used (sparse input linear, linear, crelu, screlu). you can assume people know these all you want, but you need to outline the scope. You have stuff like "2N -> 8 -> 1" without ever mentioning network composition or even linear layers in the first place, how are people supposed to understand this?
- layer types that are not used but may be used in the future (quantmoid, pooling); "[Linear layer with sparse input and blocked sparse output](https://github.com/official-stockfish/nnue-pytorch/blob/master/docs/nnue.md#linear-layer-with-sparse-input-and-blocked-sparse-output)" could go away I guess
- general principles of quantization-aware training
- (if people are still using pytorch) feature transformer kernels
- loss functions that work well

Also, at this point the history of stockfish architectures should be moved to a separate file.

PR #504, opened 2026-07-20
add architecture diagrams

sscg13:


sscg13:
Will update the dates and commits once SFNNv16 gets officially merged

PR #505, opened 2026-07-21
perf: reduce feature-transform backward memory traffic

vondele:
- Cache forward clamped activations and reuse them in backward, avoiding recomputation of the feature-transform forward pass.
- Merge L1 and PSQT passes in the fused forward/backward kernels.
- Add `__ldg` and `__restrict__` hints to improve load efficiency.

### DIFF
--- a/model/modules/feature_transformer/fused_ft_kernel.py
+++ b/model/modules/feature_transformer/fused_ft_kernel.py
@@ -25,24 +25,24 @@ typedef long long int64_t;
 
 extern "C" __global__
 void fused_double_ft_forward(
-    const float*   const us,
-    const float*   const them,
-    const int32_t* const white_indices,
-    const int32_t* const black_indices,
-    const int64_t* const psqt_indices,
-    const float*   const weight,
-    const float*   const bias,
+    const float* __restrict__ us,
+    const float* __restrict__ them,
+    const int32_t* __restrict__ white_indices,
+    const int32_t* __restrict__ black_indices,
+    const int64_t* __restrict__ psqt_indices,
+    const float* __restrict__ weight,
+    const float* __restrict__ bias,
     const float          max_ft_act,
-          float*   const l0_out,
-          float*   const wpsqt_out,
-          float*   const bpsqt_out,
+          float* __restrict__ l0_out,
+          float* __restrict__ wpsqt_out,
+          float* __restrict__ bpsqt_out,
     const int32_t        output_size
 ) {
     const uint32_t block_idx = blockIdx.x;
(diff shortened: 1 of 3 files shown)


vondele:
Consistent and fairly large speedup:

```
master noopt 1 56.56it/s
master noopt 2 56.97it/s
master noopt 3 57.30it/s
perf-gpu-ft-memory-pr noopt 1 61.32it/s
perf-gpu-ft-memory-pr noopt 2 61.52it/s
perf-gpu-ft-memory-pr noopt 3 61.51it/s
```

model/modules/feature_transformer/fused_ft_kernel.py
```diff
@@ -25,87 +25,87 @@ def make_fused_double_ft_forward_kernel(max_active_indices: int, l1_size: int):
...
+
         for(int k=0; k<""" + str(max_active_indices) + r"""; ++k) {
             int w_idx = w_idx_row[k];
             if (w_idx != -1) {
-                w0 += weight[w_idx * output_size + i];
-                w1 += weight[w_idx * output_size + i + l1_half];
+                w0 += __ldg(&weight[w_idx * output_size + i]);
+                w1 += __ldg(&weight[w_idx * output_size + i + l1_half]);
+                if (threadIdx.x == 0) {
+                    w_psqt_val += __ldg(&weight[w_idx * output_size + l1_size + p_idx]);
```
Disservin: i think this is buggy, not right now with L1 = 1024 because `output_thread_slice_size` will be 1, but for bigger L1's where it's > 1, because it will run this accumulation twice ?

Sopel97:
> Merge L1 and PSQT passes in the fused forward/backward kernels.

is this actually a positive? would be quite surprising to me

model/modules/feature_transformer/fused_ft_kernel.py
Sopel97: I think you're right, previously it was outside of the `s` loop

vondele: Indeed, let me try the fix.

vondele:
So bug-fixed and simplified, still roughly the same performance:

perf-gpu-ft-memory-pr noopt 1 60.58it/s
perf-gpu-ft-memory-pr noopt 2 60.16it/s
perf-gpu-ft-memory-pr noopt 3 60.04it/s

TonyCongqianWang:
Oh wow! Thats great!

vondele:
will close this one in favor of https://github.com/official-stockfish/nnue-pytorch/pull/508 which includes this and further improvements.

PR #506, opened 2026-07-22
Add one cycle like lamba scheduling options

TonyCongqianWang:
Used to create net for https://github.com/official-stockfish/Stockfish/pull/6993


vondele:
This will need rebasing on current master.

TonyCongqianWang:
Done!

PR #508, opened 2026-07-23
training throughput improvements (~+28% over master)

vondele:
This PR adds four performance-focused, numerically equivalent changes that together raise 4xGH200 training throughput from ~56.8 it/s to ~72.7 it/s on benchmark relevant to SF master nets.

Commit 82d53dd: Optimize fused feature-transformer memory access
  Use __ldg/__restrict__ and cache clamped activations to remove the backward re-computation of the FT forward pass.
  Measured gain: +6.5% (56.8 -> 60.46 it/s)

Commit 9ec446b: Use explicit DDPStrategy with bucket tuning
  Set gradient_as_bucket_view=True, static_graph=True, bucket_cap_mb=50, broadcast_buffers=False, and add --process_group_backend.
  Measured gain: +1.3% (60.46 -> 61.23 it/s)

Commit 58fb87e: Tile FT backward and privatize grad_bias
  Process 4 positions per CUDA block, accumulate grad_bias in shared memory, and fold PSQT accumulation into the main L1 loops.
  Measured gain: +3.5% (61.23 -> 63.39 it/s)

Commit 0ebf883: Fused CuPy RangerLite optimizer
  Replace per-parameter Python elementwise ops with fused kernels. The phase-2 path keeps the update fused even when weight-decay groups are present.
  Measured gain: +14.7% cumulative (63.39 -> 72.74 it/s)

With --factorized-weight-decay=0.001 the fused phase-2 path reaches 63.15 it/s, versus 71.24 it/s when no weight-decay groups exist. The remaining gap is the inherent cost of the global stable-weight-decay variance reduction

Verification:
- pytest tests/test_fused_double_ft.py tests/test_feature_transformer.py passes (pre-existing l1=2048 tolerance failure unchanged).
- Fused optimizer matches the Python path numerically.

### DIFF
--- a/model/modules/feature_transformer/fused_ft_kernel.py
+++ b/model/modules/feature_transformer/fused_ft_kernel.py
@@ -41,16 +41,19 @@ void fused_double_ft_forward(
 ) {
     const uint32_t block_idx = blockIdx.x;
     const uint32_t slice_offset = threadIdx.x * """ + str(output_thread_slice_size) + r""";
-    
+
     const float us_val = __ldg(&us[block_idx]);
     const float them_val = __ldg(&them[block_idx]);
-    
+
     const int32_t* const w_idx_row = white_indices + block_idx * """ + str(max_active_indices) + r""";
     const int32_t* const b_idx_row = black_indices + block_idx * """ + str(max_active_indices) + r""";
-    
+
     const int32_t l1_size = """ + str(l1_size) + r""";
     const int32_t l1_half = """ + str(l1_half) + r""";
-    
+    const int64_t p_idx = __ldg(&psqt_indices[block_idx]);
+    float w_psqt_val = __ldg(&bias[l1_size + p_idx]);
+    float b_psqt_val = __ldg(&bias[l1_size + p_idx]);
+
     #pragma unroll
     for (uint32_t s = 0; s < """ + str(output_thread_slice_size) + r"""; ++s) {
         uint32_t i = slice_offset + s;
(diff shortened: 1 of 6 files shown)


vondele:
Current nettest are
https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5137461961076608/2926829081096545/-/jobs/15483387254
https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5137461961076608/2926829081096545/-/jobs/15493519329

verifies final performance, with and without weight decay. Should soon confirm net strength.

model/modules/feature_transformer/fused_ft_kernel.py
```diff
@@ -25,87 +25,87 @@ def make_fused_double_ft_forward_kernel(max_active_indices: int, l1_size: int):
...
+
         for(int k=0; k<""" + str(max_active_indices) + r"""; ++k) {
             int w_idx = w_idx_row[k];
             if (w_idx != -1) {
-                w0 += weight[w_idx * output_size + i];
-                w1 += weight[w_idx * output_size + i + l1_half];
+                w0 += __ldg(&weight[w_idx * output_size + i]);
+                w1 += __ldg(&weight[w_idx * output_size + i + l1_half]);
+                if (threadIdx.x == 0) {
+                    w_psqt_val += __ldg(&weight[w_idx * output_size + l1_size + p_idx]);
```
43eayl5: Is it possible that `wpsqt_out` and `bpsqt_out` get over-accumulated for larger layer sizes like 2048 where `output_thread_slice_size` becomes greater than 1?

model/modules/feature_transformer/fused_ft_kernel.py
```diff
@@ -136,121 +138,128 @@ def make_fused_double_ft_backward_kernel(max_active_indices: int, l1_size: int):
...
-        
-        for(int k=0; k<""" + str(max_active_indices) + r"""; ++k) {
-            int w_idx = w_idx_row[k];
-            if (w_idx != -1) {
-                atomicAdd(&grad_weight[w_idx * output_size + i], g_w0);
-                atomicAdd(&grad_weight[w_idx * output_size + i + l1_half], g_w1);
-            } else break;
+    const int32_t tile_size = """ + str(tile_size) + r""";
+
+    __shared__ float shared_grad_bias[""" + str(l1_size + 8) + r"""];
```
43eayl5: size is fixed to `l1_size + 8` while `output_size = l1_size + num_psqt_buckets`

model/optimizers/ranger_lite.py
```diff
@@ -49,6 +251,230 @@ def __init__(
...
+        pnm_factor = group["pnm_momentum"]
+        eps = group["eps"]
+        beta1_sq = beta1 * beta1
+        one_minus_beta1_sq = 1.0 - beta1_sq
+        one_plus_pnm = 1.0 + pnm_factor
+        noise_norm = math.sqrt((1.0 + pnm_factor) ** 2 + pnm_factor ** 2)
+
+        wd_factor = 0.0
+        if decay and variance_normalized is not None:
+            wd_factor = float(decay * lr / variance_normalized.item())
```
43eayl5: (nitpick)
Could evaluate `variance_normalized.item()` once outside the group and passed in as an argument

model/modules/feature_transformer/fused_ft_kernel.py
Sopel97: https://github.com/official-stockfish/nnue-pytorch/pull/505#discussion_r3621062207

vondele: should be fixed. Thanks.

model/modules/feature_transformer/fused_ft_kernel.py
vondele: idem

model/optimizers/ranger_lite.py
vondele: idem

vondele:
First nettest shows an equivalent net trained : Elo: 0.55 +/- 1.39, nElo: 1.10 +/- 2.78

TonyCongqianWang:
Wow wow wow!!

TonyCongqianWang:
Could also try replacing stable weight decay with normal weight decay. Honestly don't know if stable weight decay does any good. Should be a lot faster

vondele:
Also the second training finished .... removing weight decay is not only faster, but also much stronger: 
Elo: 4.46 +/- 2.40, nElo: 9.01 +/- 4.84 .. this was not on the pp architecture however.

I think this goes a long way to prove the PR is correct, and could be merged. I potentially  have another follow up.

@xu-shawn opinions?

xu-shawn:
lgtm from a quick skim

xu-shawn:
do we have non-cuda fallbacks for the cuda optimizer?

vondele:
yes, cpu testing is fine.

TonyCongqianWang:
We might also try to use adam(W) again.

Removing lookahead would for the very least reduce memory by a lot and probably also speed things up

PR #509, opened 2026-07-24
Remove PyTorch Lightning from training; introduce plain SimpleTrainer

vondele:
- Replace model/lightning_module.py with a plain torch.nn.Module NNUE class.
- Add trainer/engine.py with SimpleTrainer (DDP, gradient clipping, checkpointing).
- Add trainer/callbacks.py and trainer/loggers.py for plain callbacks/logging.
- Rewrite train.py to use SimpleTrainer; compile inner NNUEModel only.
- Update serialize.py, cross_check_eval.py, ftperm.py, load_model.py and
  tests/test_training_pipeline_run.py to the new plain .ckpt format.
- Remove Lightning dependency check from scripts/easy_train.py.
- Fix CPU DDP backend selection and Gloo all-reduce for validation loss.
    
Paired 10-epoch 4xGH200 benchmark:
  perf-gpu-gh200-pr: ~71.7 it/s
  plain-trainer-rewrite: ~76.9 it/s
   (+7.2%).


vondele:
On top of https://github.com/official-stockfish/nnue-pytorch/pull/508

Nettest running https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5137461961076608/2926829081096545/-/jobs/15525911529

ruff needs fixing.

vondele:
ruff fixed, the updated ruff triggered more rules see also https://github.com/official-stockfish/nnue-pytorch/pull/510

PR #510, opened 2026-07-24
Restore Ruff Lint

Torom:
In [v0.16](https://astral.sh/blog/ruff-v0.16.0), ruff significantly expanded its default rule set.
This patch reverts that change and uses the old, very limited rule set.


vondele:
I think I'm in the process of fixing all..

Torom:
Even better

PR #511, opened 2026-07-26
slightly refactor optimizer wrappers and add AdamW wrapper

TonyCongqianWang:
Adding support for AdamW which strips away the remaining extras in rangerlite that were not ablated yet

### DIFF
--- a/model/optimizers/adamw_wrapper.py
+++ b/model/optimizers/adamw_wrapper.py
@@ -1,0 +1,88 @@
+from dataclasses import dataclass
+from typing import Any
+
+import torch
+
+from .lr_scheduler import LRSchedulerConfig, setup_lr_scheduler
+
+
+@dataclass(kw_only=True, frozen=False)
+class AdamWConfig(LRSchedulerConfig):
+    adamw_beta1: float = 0.9
+    """AdamW beta1 coefficient."""
+
+    adamw_beta2: float = 0.999
+    """AdamW beta2 coefficient."""
+
+    adamw_eps: float = 1.0e-8
+    """AdamW epsilon stability parameter."""
+
+    adamw_fused: bool = True
+    """Whether to use fused kernel for AdamW if supported by PyTorch and device."""
+
+
+class AdamWWrapper:
+    def __init__(self, config):
+        self.config = config
+        self.lr = config.lr
+        self.adamw_beta1 = config.adamw_beta1
+        self.adamw_beta2 = config.adamw_beta2
+        self.adamw_eps = config.adamw_eps
+        self.adamw_fused = config.adamw_fused
+        self.needs_train_flip = False
+        self.optimizer = None
+
+    def configure_optimizers(self, train_params):
+        betas = (self.adamw_beta1, self.adamw_beta2)
+        if self.adamw_fused:
+            try:
+                self.optimizer = torch.optim.AdamW(
+                    train_params,
+                    lr=self.lr,
+                    betas=betas,
+                    eps=self.adamw_eps,
+                    fused=True,
+                )
+            except (RuntimeError, TypeError, ValueError):
+                self.optimizer = torch.optim.AdamW(
+                    train_params,
+                    lr=self.lr,
+                    betas=betas,
+                    eps=self.adamw_eps,
+                    fused=False,
+                )
+        else:
+            self.optimizer = torch.optim.AdamW(
+                train_params,
+                lr=self.lr,
+                betas=betas,
+                eps=self.adamw_eps,
+                fused=False,
+            )
+
+        scheduler = setup_lr_scheduler(self.optimizer, train_params, self.config)
+        return [self.optimizer], [scheduler]
+
+    def switch_to_train(self, force=False):
+        pass
+
+    def switch_to_eval(self):
+        pass
+
+    def on_train_epoch_start(self, pl_module: Any):
+        pass
+
+    def on_train_batch_start(self, pl_module: Any, batch, batch_idx):
+        pass
+
+    def on_validation_epoch_start(self, pl_module: Any):
+        pass
+
+    def on_test_epoch_start(self, pl_module: Any):
+        pass
+
+    def on_train_epoch_end(self, pl_module: Any):
+        pass
+
+    def on_save_checkpoint(self, pl_module: Any, checkpoint):
+        pass
--- a/model/optimizers/config.py
+++ b/model/optimizers/config.py
@@ -1,13 +1,14 @@
 from dataclasses import dataclass
 from typing import Literal
 
+from .adamw_wrapper import AdamWConfig, AdamWWrapper
 from .rangerlite_wrapper import RangerLiteConfig, RangerLiteWrapper
 from .schedulefree_wrapper import ScheduleFreeConfig, ScheduleFreeWrapper
 
 
 @dataclass(kw_only=True)
-class OptimizerConfig(RangerLiteConfig, ScheduleFreeConfig):
-    optimizer_name: Literal["schedulefree", "ranger21", "rangerlite"] = "rangerlite"
+class OptimizerConfig(RangerLiteConfig, ScheduleFreeConfig, AdamWConfig):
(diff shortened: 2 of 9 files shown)


TonyCongqianWang:
I think this is very undisruptive (changes no defaults) while adding some decent options to play around with.

In particular it does seem as adamW could be quite promising for the future

I'd say its good to be merged?

@vondele

TonyCongqianWang:
Required for https://github.com/vondele/nettest/pull/448

vondele:
I've merged the nettest PR since that's the current master net. @xu-shawn can you have a look and merge this?

xu-shawn:
I had Claude write up a code review. Will take a look later

## Findings

**1. The AdamW fused fallback produces a broken optimizer — `model/optimizers/adamw_wrapper.py:38-53`**

`Optimizer.__init__` calls `add_param_group`, which stamps defaults into *the caller's* group dicts, and `Adam.__init__` validates `fused` only *after* that. So the failed `fused=True` attempt leaves `fused: True` inside `train_params`. The retry uses `setdefault`, so `fused=False` loses to the stale value:

```
after failed ctor: {'lr': 0.001, 'weight_decay': 0.01, ..., 'fused': True}
defaults.fused = False | group.fused = True
STEP FAILED: NotImplementedError Could not run 'aten::_fused_adamw_' with arguments from the 'CPU' backend
```

The fallback constructs fine, so nothing complains until the first `optimizer.step()`, which dies with a confusing dispatcher error. Reproduced on your torch 2.3.1+rocm5.7 with plain CPU params. CI won't catch it: torch 2.8 supports fused on CPU so the fallback never fires there, and no test in the new file ever calls `.step()` on the AdamW optimizer.

Fix is to not reuse the mutated dicts — either `copy.deepcopy(train_params)` for the retry, or set `fused=False` explicitly in each group before retrying, or probe device support up front instead of try/except.

**2. `weight_decay` is never passed to `torch.optim.AdamW` — `adamw_wrapper.py:39-53`**

AdamW's own default is `0.01`, so any group omitting `weight_decay` silently gets decay while every other optimizer here defaults to `0.0`. Confirmed: `configure_optimizers([{"params":[r], "lr":1e-3}])` → `group.weight_decay = 0.01`. Harmless today only because `model/nnue.py:127-179` sets it on all ten groups — but `RangerLiteWrapper` deliberately passes `weight_decay=0.0` as the fallback with a comment saying so (`rangerlite_wrapper.py:49`). Worth matching.

**3. Dead work in the slow path — `ranger_lite.py:614, 654-655`**

Phase 1 still unconditionally does `variance_ma / bias_correction2` (a full tensor allocation per parameter per step) and accumulates the sum, but with the new gate at line 658 the result is discarded whenever `use_stable_weight_decay=False` and normloss is off. The `needs_variance_sum` flag is right there — guarding the two lines with it removes a per-step allocation from every non-stable-decay CPU/no-cupy run.

**4. The decay test can't fail — `tests/test_optimizers.py:206`**

`test_ranger_stable_weight_decay_option` ends at `assert not torch.allclose(param_ew, initial_p)`, which the Adam step alone satisfies; the test passes unchanged if elementwise decay is never applied. Since that flag is the point of the commit, it's worth asserting the actual value — e.g. with `lr=0.1, wd=0.5` on `p=1.0` and a zero-ish gradient, or just asserting `stable` and `non-stable` produce different results from `wd=0`. Adding a `.step()` to the AdamW tests would also have caught finding 1.

**5. Minor**

- The fused fallback is silent (`adamw_wrapper.py:46`), so a run that quietly loses the fused kernel looks identical to one that keeps it — a warning would help, and the `except` is broad enough to swallow genuine config errors.
- `model/__init__.py` re-exports `AdamWConfig`/`AdamWWrapper` but not the equally new `LRSchedulerConfig`/`SafeOneCycleLR`, which `model/optimizers/__init__.py` does export.

The `LRSchedulerConfig` extraction itself is clean — the MRO `OptimizerConfig → RangerLiteConfig → ScheduleFreeConfig → AdamWConfig → LRSchedulerConfig` linearizes fine, `trainer/engine.py:53-63` still detects `SafeOneCycleLR` by class name after the move, and ruff 0.16.0 passes on all changed files.

vondele:
since we're now using this quite extensively, getting this merged would be good.

TonyCongqianWang:
Shawn said he is a bit busy right now... So maybe someone else review and I fix the claude comments together with whatever get commented too?

model/optimizers/adamw_wrapper.py
```diff
@@ -0,0 +1,88 @@
...
+    def configure_optimizers(self, train_params):
+        betas = (self.adamw_beta1, self.adamw_beta2)
+        if self.adamw_fused:
+            try:
+                self.optimizer = torch.optim.AdamW(
+                    train_params,
+                    lr=self.lr,
+                    betas=betas,
+                    eps=self.adamw_eps,
+                    fused=True,
```
xu-shawn: I don't think we need this exception handling. passing in `fused=self.adamw_fused` should suffice?

model/optimizers/config.py
```diff
@@ -1,13 +1,14 @@
...
+from .adamw_wrapper import AdamWConfig, AdamWWrapper
 from .rangerlite_wrapper import RangerLiteConfig, RangerLiteWrapper
 from .schedulefree_wrapper import ScheduleFreeConfig, ScheduleFreeWrapper
 
 
 @dataclass(kw_only=True)
-class OptimizerConfig(RangerLiteConfig, ScheduleFreeConfig):
-    optimizer_name: Literal["schedulefree", "ranger21", "rangerlite"] = "rangerlite"
+class OptimizerConfig(RangerLiteConfig, ScheduleFreeConfig, AdamWConfig):
+    optimizer_name: Literal["schedulefree", "ranger21", "rangerlite", "adamw"] = "rangerlite"
```
xu-shawn, line 18: can we use a {name: class} dict here? that would simplify the lookup in get_optimizer_wrapper and the exception message as well

model/optimizers/lr_scheduler.py
```diff
@@ -0,0 +1,53 @@
...
+
+class SafeOneCycleLR(torch.optim.lr_scheduler.OneCycleLR):
+    def step(self, epoch=None):
+        if self.last_epoch < self.total_steps - 1:
+            super().step(epoch)
+
+
+def setup_lr_scheduler(
+    optimizer: torch.optim.Optimizer,
+    train_params: list[dict[str, Any]],
```
xu-shawn: can we just derive this field from `optimizer.param_groups`

model/optimizers/rangerlite_wrapper.py
```diff
@@ -34,26 +20,22 @@ class RangerLiteConfig:
...
+
 
 class RangerLiteWrapper:
     def __init__(
         self,
         config,
         legacy_mode,
     ):
-        self.gamma = config.gamma
+        self.config = config
```
xu-shawn, line 33: since we are storing config anyway, we don't need to copy over the fields in subsequent lines anymore

model/optimizers/ranger_lite.py
```diff
@@ -674,9 +701,12 @@ def step(self, closure=None):
 
                 # --- CORRECT BEHAVIOR ---
                 if not self.use_legacy_scoping_bug:
-                    # Stable Weight Decay
+                    # Weight Decay
                     if decay:
-                        p.data.mul_(1 - decay * lr / variance_normalized)
+                        if self.use_stable_weight_decay and variance_normalized is not None:
```
xu-shawn: can you put the two decay logic into a single function. Claude also pointed out different behavior between CUDA & PyTorch implementations when `self.use_stable_weight_decay` is true but `variance_normalized` is None. could you add an assert to prevent that?

TonyCongqianWang:
I tried addressing the comments as best as I could. Since I cant test locally, please test and rereview!

vondele:
I ran a few iterations of our production settings with this, and that didn't expose a problem.

vondele:
ran a full training based on this branch and with https://github.com/official-stockfish/nnue-pytorch/pull/516 on top, and that produced a master quality net, so this looks OK.

